PR #6, opened 2014-10-01
Make the hash algorithm configurable per DB

sagelywizard:


kxepal:
What the other than `crc32` algorithms are available?

sagelywizard:
None currently. This change and https://github.com/apache/couchdb-mem3/pull/3 just makes it easy to add others.

PR #8, opened 2014-11-07
2080 port cors

chewbranca:
This brings in support for CORS into `chttpd` itself. It is based off of the `couch_httpd_cors` implementation and is a sizable refactor that cleans up a number of things and also makes the core functionality completely pure. The only IO is performed in `chttpd_cors:maybe_handle_preflight_request/1` and `chttpd_cors:headers/2` which both call `get_cors_config/1`. The function `get_cors_config` loads the server wide CORS settings and then passes that through the rest of the module. We can then test all the functionality of the module directly without needing to do any test bed server loading or module level mocking.

### DIFF
--- a/include/chttpd_cors.hrl
+++ b/include/chttpd_cors.hrl
@@ -1,0 +1,81 @@
+% Licensed under the Apache License, Version 2.0 (the "License"); you may not
+% use this file except in compliance with the License. You may obtain a copy of
+% the License at
+%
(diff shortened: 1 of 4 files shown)


kxepal:
Hey @chewbranca! What's the status?

kxepal:
friendly nudge @chewbranca (:

rnewson:
+1

PR #9, opened 2014-11-19
Return ok:true on _bulk_doc update

robertkowalski:
Return `ok: true` on a successful _bulk_doc update like in
CouchDB 1.6.

Example success message on 2.0:
```
[{"id":"Brocket","rev":"6-b6fa9e703c3eb92aa9c3a49cedf1e8c2"}]
```

Example error message on 2.0:
```
[{"id":"Blaggie-System","error":"conflict",
  "reason":"Document update conflict."}]
```

Finally closes COUCHDB-2462

### DIFF
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -840,9 +840,9 @@ update_doc_result_to_json({{Id, Rev}, Error}) ->
 update_doc_result_to_json(#doc{id=DocId}, Result) ->
     update_doc_result_to_json(DocId, Result);
 update_doc_result_to_json(DocId, {ok, NewRev}) ->
-    {[{id, DocId}, {rev, couch_doc:rev_to_str(NewRev)}]};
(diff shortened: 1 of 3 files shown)


kxepal:
Hey @robertkowalski ! What's the status?

kxepal:
Very not sure about test suite - we should have all the helpers to not deal with those routes everywhere. However, that's  a matter of separate work which will be done in anyway.

robertkowalski:
I agree! So far I don't have a better solution and at least we have some tests that will detect if the regression occurs again and they also confirm in an automatic way that the bug is fixed.

robertkowalski:
Can we get this bugfix merged?

I know that you are working on a new way for a better testsuite, but I think it should not block fixing this issue + adding regression tests. There is also a main difference for the testhelper compared to the ones in couchdb-couch: I additionally start `chttpd`

kxepal:
+1. merge it (:

robertkowalski:
merged as 4ea930e62f4e2fceb0e13a62ecd90cb628477c9c - weeee!

PR #10, opened 2014-11-28
Dynamic endpoints handlers

hdiedrich:
chttpd hardcoded handlers are replaced with with dynamic url handlers,
which are functions in a dynamically created module, assembled from
priv/chttpd_handler.cfg files from all interested applications.

For couched these are currently: chttpd, mem3 and global_changes.

This is a special branch that pull the prepared branches (of the same name)
of chttpd, mem3 and global_changes, which have the config files.

Run 'make check' to test the handlers. They are implicitly tested by other
tests and there is a whitebox callback test chttpd_handler_callback_test,
which calls the dynamic functions directly and mocks the funs that they
return. This tests precisely the relationship between the *_handler function
clauses and the returned function.

There are more tests to come, both testing the endpoints via http, and the
configuration assembly.

BugzID: 27037

### DIFF
--- a/test/chttpd_handler_callback_test.erl
+++ b/test/chttpd_handler_callback_test.erl
@@ -31,179 +31,179 @@
 
 -include_lib("eunit/include/eunit.hrl").
 
+
 all_test_() ->
-    io:format(user, "~nEndpoint handler callbacks:~n", []),
-    {setup,
+    {foreach,
     fun() -> chttpd_handler:build(test_cfg()) end,
 
     % url handlers
-    [fun test_empty_string/0,
-    fun test_favicon/0,
-    fun test_utils/0,
-    fun test_all_dbs/0,
-    fun test_active_tasks/0,
-    fun test_config/0,
-    fun test_reload_query_servers/0,
-    fun test_replicate/0,
-    fun test_uuids/0,
-    fun test_sleep/0,
-    fun test_session/0,
-    fun test_oauth/0,
-    fun test_up/0,
-    fun test_membership/0,
-    fun test_db_updates/0,
-    fun test_anything/0,
+    [fun test_empty_string/1,
+    fun test_favicon/1,
+    fun test_utils/1,
+    fun test_all_dbs/1,
+    fun test_active_tasks/1,
+    fun test_config/1,
+    fun test_reload_query_servers/1,
+    fun test_replicate/1,
+    fun test_uuids/1,
+    fun test_sleep/1,
+    fun test_session/1,
+    fun test_oauth/1,
+    fun test_up/1,
+    fun test_membership/1,
+    fun test_db_updates/1,
+    fun test_anything/1,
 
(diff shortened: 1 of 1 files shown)


kxepal:
I'm happy with merge this. @rnewson @davisp @chewbranca what do you think?

src/chttpd_handler.erl
```diff
@@ -0,0 +1,202 @@
...
+    {design_url_handlers, list, lists:flatten(AllDesignHandlers)}].
+
+-spec load_defs(AppName::atom()) -> OneAppsHandlerCfg::config().
+%% @doc assemble the configuration from the chttpd_handler.cfg of all apps.
+load_defs(App) ->
+    case code:priv_dir(App) of
+        {error, _Error} ->
+            [];
+        Dir ->
+            Path = Dir ++ "/chttpd_handler.cfg",
```
kxepal: Seems I miss that moment. Why not to reuse `config` and have all the handlers defined there? We already keep httpd ones there and this gives us dynamic control over them with no server restart.

iilyak:
@hdiedrich: Any updates on this PR?

hdiedrich:
Hi, I need to review what that over all situation for this is at this point. Status is that there was a wish from Paul on how to better kick the handlers in, which I liked, but got stuck implementing it and kept going backwards finding out why until it became clear that I need to port my system tests over from the original implementation in dbcore that had met little love there. I then waited for integration tests from Russel and Ilya to be merged, so I don't double their effort but those got stuck, too IIRC. The way forward should be simple and fast and I'll be happy to finally get it merged.

The start up, as it is, has the weakness that there needs to be knowledge about the modules in play, which defeats the intended uncoupling. The solution would be to just re-compile every step of the way when a new app is added during start up. That makes it fool proof and self-sufficient. Config, as Alex suggests, should therefore not be needed.

kxepal:
@hdiedrich Hey! Config support suddenly is needed by the following reasons:

- It saved us many times when security issue was found in some HTTP handler. The recent example: http://docs.couchdb.org/en/latest/cve/2014-2668.html - users were able to handle that problem without upgrade procedure, but with a single config line change.

- Some users are disables default handlers, or provide own overlay, in order to prevent or customize access to certain endpoints. Quite popular move is to disable /_all_dbs in order to prevent sharing database lists for everyone. Another option is to replace default handler with own from the plugin if you're going to change behaviour without patching the source code.

- Somehow need to provide ability to register proxy and external handlers.

rnewson:
The plan that the couchdb 2.0 sprint team agreed on is basically;

1) every application can easily register its own endpoints simply by existing
2) administrators can disable any endpoints they like
3) the configuration file's proxy and external handlers section should be included in the set of endpoints (and added last)

As to how we can support 'overlay' of an application's endpoints, I'm less sure. I suggest we get the above steps done first and then reassess.

kxepal:
Dynamic handlers were landed based on @iilyak work. Could you close this PR please?

PR #12, opened 2014-12-04
Add users db security rules on clustered interface

mikewallace1979:
Apply authentication_db security rules when authentication_db is on the clustered interface (5984).

COUCHDB-2452

### DIFF
--- a/src/chttpd_auth_cache.erl
+++ b/src/chttpd_auth_cache.erl
@@ -12,12 +12,14 @@
 
 -module(chttpd_auth_cache).
 -behaviour(gen_server).
+-behaviour(config_listener).
 
 -export([start_link/0, get_user_creds/1]).
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2,
 	 code_change/3]).
 -export([listen_for_changes/1, changes_callback/2]).
 -export([update_auth_doc/1]).
+-export([handle_config_change/5]).
 
 -include_lib("couch/include/couch_db.hrl").
 
@@ -76,8 +78,12 @@ get_from_cache(UserName) ->
 %% gen_server callbacks
 
 init([]) ->
+    ok = config:listen_for_changes(?MODULE, nil),
     {ok, #state{changes_pid = spawn_changes(0)}}.
 
+handle_call(restart_listener, _From, #state{changes_pid=Pid} = State) ->
+    exit(Pid, kill),
+    {reply, ok, State};
 handle_call(_Call, _From, State) ->
     {noreply, State}.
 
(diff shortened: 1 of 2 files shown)


src/chttpd_db.erl
```diff
@@ -248,9 +264,21 @@ delete_db_req(#httpd{}=Req, DbName) ->
         throw(Error)
     end.
 
+get_db_options(DbName) ->
+    IsReplicatorDb = DbName == config:get("replicator", "db", "_replicator"),
+    IsUsersDb = DbName ==config:get("chttpd_auth", "authentication_db", "_users") orelse
+    binary_to_list(mem3:dbname(DbName)) == config:get("chttpd_auth", "authentication_db", "_users"),
```
kxepal: This definitely need to be simplified: no need to get the same config value twice. Also, without indention this a bit confuses.

src/chttpd_auth_cache.erl
```diff
@@ -91,6 +102,12 @@ handle_info({'DOWN', _, _, Pid, Reason}, #state{changes_pid=Pid} = State) ->
     {noreply, State#state{last_seq=Seq}};
 handle_info({start_listener, Seq}, State) ->
     {noreply, State#state{changes_pid = spawn_changes(Seq)}};
+handle_info({gen_event_EXIT, {config_listener, ?MODULE}, _Reason}, State) ->
+    erlang:send_after(5000, self(), restart_config_listener),
```
kxepal: Why 5000?

mikewallace1979: I used the same value as for restarting the config listener over in https://github.com/apache/couchdb-couch/blob/master/src/couch_auth_cache.erl#L235 - I don't know exactly why 5000 was used there, I'd guess possibly to limit the frequency in the event of repeated gen_event_EXIT messages?

src/chttpd_db.erl
mikewallace1979: Yes this line is quite horrible. I'll figure out a better way.

src/chttpd_auth_cache.erl
kxepal: May be. I'm ok with, just wondering about magic numbers (:

src/chttpd_db.erl
mikewallace1979: @kxepal I was looking at tidying this up and realised I made two mistakes here. Firstly, there is no need for the `binary_to_list(mem3:dbname(DbName))` comparison because that would only be necessary if DbName was for an individual shard - that will never be the case here so that check can be removed.

Secondly, there is no need to check for the replicator DB here because chttpd does nothing with that information.

I'll add some commits to fix those things up.

src/chttpd_auth_cache.erl
davisp: The 5000 is to limit restart frequency which can cause badness in its own right.

davisp:
+1 to squerge

kxepal:
Related:
https://github.com/apache/couchdb/pull/288
https://github.com/apache/couchdb-couch-mrview/pull/11
https://github.com/apache/couchdb-couch/pull/20

PR #18, opened 2015-01-15
Convert {timeout, Error} errors to timeout

mikewallace1979:
If chttpd got a {timeout, Error} error then it would return
that tuple to the client. Since Error is some internal specifics
this isn't particularly useful information to return to the client.

This commit converts it to our usual timeout response.

COUCHDB-2425

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -753,6 +753,8 @@ error_info(not_implemented) ->
 error_info(timeout) ->
     {500, <<"timeout">>, <<"The request could not be processed in a reasonable"
         " amount of time.">>};
+error_info({timeout, _Reason}) ->
+    error_info(timeout);
 error_info({Error, null}) ->
     error_info(Error);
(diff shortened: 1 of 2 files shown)


mikewallace1979:
nvm

kxepal:
What comes with the `_Reason`? May be it's worth to preserve and pop up?

mikewallace1979:
Most likely a stack trace. We could preserve it instead - I'm happy either way.

kxepal:
Not good then. Let it remains in logs.

kxepal:
Patch needs to be rebased + please add a [test case](https://github.com/apache/couchdb-chttpd/blob/master/test/chttpd_error_info_tests.erl#L132-L137) (it's trivial) about to be sure that it wouldn't get accidentally be broken.

mikewallace1979:
@kxepal all done - good call on the test case

kxepal:
Awesome! Merged, thank you!

PR #19, opened 2015-01-29
validate _all_docs query parameters

willholley:
Perform the same query parameter validation in the clustered interface for /_all_docs requests as we do in the non-clustered interface. Fixes COUCHDB-2523.


kxepal:
+1

kxepal:
Merged in 9a8e650997 thanks!

PR #20, opened 2015-01-29
Set view type before validating _all_docs parameters

willholley:
Matching couch_mrview, set view_type to "map" before validating query parameters for _all_docs. This fixes a bug whereby validation fails when specifying _all_docs?conflicts=true. I missed this in the previous fix for COUCHDB-2523.


kxepal:
that's why adding tests on changes is a good idea (;

kxepal:
Thanks!

PR #25, opened 2015-02-19
Remove _config route on cluster

robertkowalski:
In order to avoid users shooting themselves in the foot by using
`/_config/` on a clustered CouchDB with a loadbalancer in front,
we remove it on `15984` - it will be available for single-node-
mode on the backdoor port (`15986`) or for users that are feeling
lucky which want to fire curl requests to every node.

It also allows Fauxton to detect if it is running on a the backdoor
port. Fauxton will - if it gets a 200 instead of a 404 - show the
config-section to the user.

COUCHDB-2601 COUCHDB-2390 COUCHDB-2343


robertkowalski:
cc @benkeen

janl:
screen shot 2015-02-19 at 22 18 42
+1

PR #26, opened 2015-02-25
Return {error, Reason} for illegal_database_name

iilyak:
COUCHDB-2625

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -745,10 +745,11 @@ error_info({bad_ctype, Reason}) ->
     {415, <<"bad_content_type">>, Reason};
 error_info(requested_range_not_satisfiable) ->
     {416, <<"requested_range_not_satisfiable">>, <<"Requested range not satisfiable">>};
-error_info({error, illegal_database_name}) ->
-    {400, <<"illegal_database_name">>, <<"Only lowercase letters (a-z), "
-        "digits (0-9), and any of the characters _, $, (, ), +, -, and / are "
-        "allowed. Moreover, the database name must begin with a letter.">>};
+error_info({error, {illegal_database_name, Name}}) ->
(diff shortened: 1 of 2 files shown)


davisp:
The PR tile, linked ticket, and diff don't seem to match to me. The diff appears to change the error message to include the database name (which is fine, but I don't think we should change the text, just append "Provided name was 'foo'" or something).

It doesn't return {error, Reason} anywhere that I can see which the commit and PR title suggest this change does.

The linked ticket is about changing things to be able to configure a database name callback which doesn't have anything at all to do with this ticket.

iilyak:
Here is the reason for the change https://github.com/iilyak/couchdb-couch/commit/0e01a33cc6443f0690cc5f2460c0240b64eda335. However I did close the main PR. The rationale for changing message text was discussed in closed PR I linked above. 

> @kxepal: In case of db name validation customization this error message shouldn't be defined here since the requirements for the name could be different.

Another orphan PR related to the issue: https://github.com/apache/couchdb-fabric/pull/14

The code duplication need to be eliminated anyway regardless of configurable dbname validator feature.
Also I do believe all errors need to be `{error, Reason}` across the system. Since it allows you to wrap functions when you need it without knowing all possible errors that wrapped function could return. You just handle `{error, Reason}` do error case logic and bubble `{error, Reason}` up.   

I can create a new main PR which would just implement couch_db:validate_dbname and changes the error `{error, illegal_database_name, DbName}` -> `{error, {illegal_database_name, DbName}}`. I could create new jira issue with following description: `Remove code duplication in dbname validation`. If I go with second approach I would need to rewrite history for this PR to refer to different jira issue. I could create 2 different JIRA issues one for {error, Reason} and another one for code duplication.

Please advise what would be the better (I prefer second).

iilyak:
stale PR

PR #27, opened 2015-02-25
Rename system databases

kxepal:


chewbranca:
Looks like there's a discrepancy on the naming of the config value. The subfield should be `shards_db` (plural) not `shard_db`. I'm treating the plural version as canonical because that's what [mem3_shards.erl](https://github.com/apache/couchdb-mem3/blob/master/src/mem3_shards.erl#L228-L229) uses when it creates the database.

kxepal:
@chewbranca nice found! fixed it in https://github.com/kxepal/couchdb-mem3/commit/b727f2d969040040357c06b3cf7f7158b42f7bac for https://github.com/apache/couchdb-mem3/pull/6

PR #28, opened 2015-03-01
add _changes?feed=stream sugar for continuous

robertkowalski:
allow `feed=stream` as sugar for `continuous` which is hard to
type.

PRs for the change:
https://github.com/apache/couchdb/pull/307
https://github.com/apache/couchdb-couch/pull/40
https://github.com/apache/couchdb-chttpd/pull/28

closes COUCHDB-2237

### DIFF
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -93,7 +93,7 @@ handle_changes_req1(#httpd{}=Req, Db) ->
             couch_stats:decrement_counter([couchdb, httpd, clients_requesting_changes])
         end;
     _ ->
-        Msg = <<"Supported `feed` types: normal, continuous, longpoll, eventsource">>,
+        Msg = <<"Supported `feed` types: normal, continuous, live, longpoll, eventsource">>,
         throw({bad_request, Msg})
     end.
 
@@ -1314,6 +1314,9 @@ parse_changes_query(Req) ->
     erlang:erase(changes_seq_interval),
     ChangesArgs = lists:foldl(fun({Key, Value}, Args) ->
         case {string:to_lower(Key), Value} of
+        {"feed", "live"} ->
+            %% sugar for continuous
+            Args#changes_args{feed="continuous"};
         {"feed", _} ->
             Args#changes_args{feed=Value};
         {"descending", "true"} ->
--- a/test/chttpd_db_test.erl
+++ b/test/chttpd_db_test.erl
@@ -31,6 +31,11 @@ create_db(Url) ->
     {ok, _, _, _} = test_request:put(Url,
                 [{"Content-Type", "application/json"}], "{}").
 
+
+create_doc(Url, Id) ->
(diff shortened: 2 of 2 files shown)


src/chttpd_db.erl
```diff
@@ -93,7 +93,7 @@ handle_changes_req1(#httpd{}=Req, Db) ->
             couch_stats:decrement_counter([couchdb, httpd, clients_requesting_changes])
         end;
     _ ->
-        Msg = <<"Supported `feed` types: normal, continuous, longpoll, eventsource">>,
+        Msg = <<"Supported `feed` types: normal, continuous/stream, longpoll, eventsource">>,
```
kxepal: One thought just crossed my mind: don't you think, that `continuous/stream` could be misread by a user and they'll try to send a request with `?feed=continuous/stream` parameter?

robertkowalski: in general a person can misread every information you present and in this case i would doubt that this happens often

how would you present the message?

robertkowalski: @daleharvey what do you think?

kxepal: `Supported `feed` types: normal, continuous, stream, longpoll, eventsource` e.g. without making difference between continuous and stream since both are the valid options. All the questions "why" will be covered by the docs.

rnewson: I agree with @kxepal that `stream` should be listed like the other options.

I'm -0 on adding an alias for an existing feature (that is, I'd rather we didn't add more api without adding functionality, but I'm not blocking it).

janl: I’m with @rnewson.

janl: a bit of clarification, I most care about the error message being clear and I suggest we use @kxepal’s version, if we add the alias. I don’t advocate for or against the alias, but if others think, that it is a good idea, you should go ahead.

davisp:
Voting 0 with no modifier as per same reason as @janl and @rnewson. Also agree using a comma instead of the slash when listing the possible values.

robertkowalski:
thank you all for the input. fixed the error message. mind taking a second look @davisp?

robertkowalski:
Hum, I changed the error message, but as nobody is really a fan of it (so far -0 and 0) and I am also unsure about it I am going to close this and will give Dale feedback. 

One thing left is that replication still uses continuous and stream would not be a good name for it. So I think we need to think about it more.

Maybe we can change `continuous` in version 3 as a breaking change for both (changes feed and replication).

Thanks for all your feedback!

kxepal:
@robertkowalski ping (:

rnewson:
feed=live already landed in couchdb-chttpd master, shouldn't this be closed now?

robertkowalski:
thanks for the reminder - merged as ca0195f38645b770c63715022ca14b7cdf649208

PR #32, opened 2015-04-13
Remove /_sleep endpoint

kxepal:
I was thought that only Windows command line needs a ping-sleep hack. Don't think we need this one.


janl:
FINALLY \o/

PR #34, opened 2015-04-29
Handle Etag generation in one place

iilyak:
COUCHDB-2650

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -19,7 +19,8 @@
     qs_value/3, qs/1, qs_json_value/3, path/1, absolute_uri/2, body_length/1,
     verify_is_server_admin/1, unquote/1, quote/1, recv/2, recv_chunked/4,
     error_info/1, parse_form/1, json_body/1, json_body_obj/1, body/1,
-    rev_etag/1, doc_etag/1, make_etag/0, make_etag/1, etag_respond/3, etag_match/2,
+    rev_etag/1, doc_etag/1, make_etag/0, make_etag/1,
+    att_etag/2, etag_respond/3, etag_match/2,
     partition/1, serve_file/3, serve_file/4,
     server_header/0, start_chunked_response/3,send_chunk/2,
     start_response_length/4, send/2, start_json_response/2,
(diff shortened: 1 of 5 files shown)


kxepal:
Good idea! May be also extract attachment etag generation for consistency?

src/chttpd.erl
```diff
@@ -535,7 +535,15 @@ json_body_obj(Httpd) ->
...
 
 doc_etag(#doc{revs={Start, [DiskRev|_]}}) ->
-    "\"" ++ ?b2l(couch_doc:rev_to_str({Start, DiskRev})) ++ "\"".
+    rev_etag({Start, DiskRev}).
+
+rev_etag({Start, DiskRev}) ->
+    Rev = couch_doc:rev_to_str({Start, DiskRev}),
+    <<"\"", Rev/binary, "\"">>.
+
+make_etag() ->
```
rnewson: a random uuid is essentially useless. The only reason cloudant does that (for views) is to preserve the shape of the code relative to couch. Better to omit the etag header entirely if it isn't going to be stable.

kxepal: Why it was used in CouchDB then? Not because to overcome some cache issues?

iilyak:
Should I call `chttpd`'s version of etag related functions from `couch` application in cases [like](https://github.com/apache/couchdb-couch/blob/952a41c7a10fc4dcb0324d09c5be7e5d1d548803/src/couch_httpd_db.erl#L121). Alternatively I can update `couch_chttpd` to dispatch the call to `chttpd`. Which of the above you would prefer?

kxepal:
I think not if only you're going to slowly replace couch_httpd by chttpd. Otherwise, it will make more complicated references between subprojects (they are already quite complicated).

src/chttpd.erl
iilyak: @rnewson: I don't think changing how `views`  handle etag generation is in the scope of the bug. I could work on it in the context of another bug if you think it needs to be done before 2.0 release. I would approach this by putting `#mrst.sig` into meta. Then when we get [here](https://github.com/apache/couchdb-couch-mrview/blob/master/src/couch_mrview_http.erl#L331), we can use `#mrst.sig` to calculate etag instead of relying on `#vacc.etag` (which is uuid at this point).

davisp:
This looks mostly good to me. I'm confused by the changes to add header manipulation functions in 9ab5b83 which don't appear to be used anywhere. Is there a corresponding branch in couch_mrview that's not linked here?

davisp:
Ah, found it. Reviewing the others as well.

iilyak:
stale PR

PR #35, opened 2015-05-08
Escape database name on posts to _replicate

iilyak:
Currently we convert local references of source/target specified in
JSON body into URLs.
This commit escapes database name. So slashes in name could be used.

COUCHDB-2666

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -305,7 +305,7 @@ fix_uri(Req, Props, Type) ->
         true ->
             Props;
         false ->
-            Uri = make_uri(Req,replication_uri(Type, Props)),
+            Uri = make_uri(Req, Uri0),
             [{Type,Uri}|proplists:delete(Type,Props)]
         end
     end.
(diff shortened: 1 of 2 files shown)


src/chttpd.erl
```diff
@@ -315,9 +315,13 @@ replication_uri(Type, PostProps) ->
     {Props} ->
         couch_util:get_value(<<"url">>, Props);
     Else ->
-        Else
+        escape_dbname(Else)
     end.
 
+escape_dbname(DbName) ->
+    re:replace(DbName, "/", "%2f", [global, {return, binary}]).
```
kxepal: Why not to use `chttpd:quote` here?

kxepal:
It would be also good to have some test on this. Again, this issue was found by third party project.

hdiedrich:
LGTM -- however, it will also encode any other special characters besides '/'. Is that intended?

Mochiweb only lets unchanched:
```
-define(QS_SAFE(C), ((C >= $a andalso C =< $z) orelse
		     (C >= $A andalso C =< $Z) orelse
		     (C >= $0 andalso C =< $9) orelse
		     (C =:= ?FULLSTOP orelse C =:= $- orelse C =:= $~ orelse
		      C =:= $_))).
```

kxepal:
@hdiedrich I think it's good to use general escape logic. Otherwise, when/if we'll decide to allow non-latin database names we'll not fight with escaping over again.

+1

iilyak:
@kxepal: I'll revert it back to be `quote(Else)`.

rnewson:
@kxepal thoughts?

kxepal:
All tests passed. +1 after squash.

iilyak:
This is fixed already

PR #37, opened 2015-06-26
Don't allow security updates during cassim migration

chewbranca:
Companion PR for `cassim` manual security doc migrations.


rnewson:
+1 after addressing that one comment.

PR #38, opened 2015-06-26
Use an internal buffer to increase _changes throughput

kocolosk:
This set of commits refactors the `changes_callback` function to keep an internal data buffer and flush it out the socket after a configurable number of bytes have been accumulated. It also adds support for a new callback message that will cause it to flush the buffer at the end of each traversal of the sequence tree for "live" feeds.

COUCHDB-2724

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -35,7 +35,8 @@
     get_delayed_req/1]).
 
 -export([
-    chunked_response_buffer_size/0
+    chunked_response_buffer_size/0,
+    close_delayed_json_object/4
 ]).
 
 -record(delayed_resp, {
@@ -727,6 +728,14 @@ send_delayed_error(#delayed_resp{req=Req,resp=nil}=DelayedResp, Reason) ->
 send_delayed_error(#delayed_resp{resp=Resp}, Reason) ->
     log_error_with_stack_trace(Reason),
     throw({http_abort, Resp, Reason}).
+
+close_delayed_json_object(Resp, Buffer, Terminator, 0) ->
+    % Use a separate chunk to close the streamed array to maintain strict
+    % compatibility with earlier versions. See COUCHDB-2724
+    {ok, R1} = chttpd:send_delayed_chunk(Resp, Buffer),
+    send_delayed_chunk(R1, Terminator);
+close_delayed_json_object(Resp, Buffer, Terminator, _Threshold) ->
+    send_delayed_chunk(Resp, [Buffer | Terminator]).
 
 end_delayed_json_response(#delayed_resp{}=DelayedResp) ->
     {ok, #delayed_resp{resp=Resp}} =
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -34,6 +34,10 @@
     send_delayed_error/2, end_delayed_json_response/1,
     get_delayed_req/1]).
 
+-export([
+    chunked_response_buffer_size/0
+]).
+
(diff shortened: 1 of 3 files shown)


kxepal:
Nice! How much does this improves throughput?

src/chttpd_db.erl
```diff
@@ -74,6 +86,8 @@ handle_changes_req1(#httpd{}=Req, Db) ->
     ChangesArgs = Args0#changes_args{
         filter_fun = couch_changes:configure_filter(Raw, Style, Req, Db)
     },
+    % Default to ~filling the payload of a standard Ethernet frame
+    Max = config:get_integer("httpd", "chunked_response_buffer", 1490),
```
kxepal: I believe, this option we should advice to tweak if user iface is configured for some other MTU, right? Also, standard Ethernet frame is 1500, why 1490 had chosen here?

kocolosk: We should figure out what the optimal size really is; filling an Ethernet frame was almost an arbitrary choice on my part. We did see a larger increase in throughput with a 64KB buffer. If the frame size really does turn out to be a good target then yes, we should either programmatically discover the MTU or document the way to optimize this.

I chose 1490 instead of 1500 because mochiweb needs a couple of bytes to declare the size of the chunk and to add a newline or two. I didn't compute exactly how many bytes are required for that "envelope"; it's probably a bit less than 10 but I figured it was better to err on the conservative side than to create a new, nearly-empty packet.

kocolosk:
The speedup depends on how CPU-constrained you are; as I mentioned in the JIRA I constructed a test on my laptop where setting the buffer to 64KB resulted in about a 60% improvement in throughput over master.

kocolosk:
As I happened to be in this part of the codebase I took the opportunity to remove the `is_old_couch` UA-sniffing hack. It means that a replication mediated by a server running a CouchDB older than version 1.0.1 will not be able to use 2.0 as a source. Mediating the replication on the 2.0 box should still work, though honestly I don't know that anyone is actively testing this part of the compatibility matrix.

src/chttpd.erl
```diff
@@ -994,3 +1007,15 @@ stack_hash(Stack) ->
...
 with_default(undefined, Default) -> Default;
 with_default(Value, _) -> Value.
+
+%% @doc CouchDB uses a chunked transfer-encoding to stream responses to
+%% _all_docs, _changes, _view and other similar requests. This configuration
+%% value sets the maximum size of a chunk; the system will buffer rows in the
+%% response until it reaches this threshold and then send all the rows in one
+%% chunk to improve network efficiency. The default value is chosen so that
+%% the assembled chunk fits into the default Ethernet frame size (some reserved
+%% padding is necessary to accomodate the reporting of the chunk length). Set
```
kxepal: Typo in "accommodate"?

kxepal:
@kocolosk I wonder who still uses pre 1.0 CouchDB, but if you're interested in results, I have such on vm backups to test with 2.0.

kocolosk:
@kxepal appreciate the offer, but personally I'd be more interested in validating replication interop as an extended part of our CI setup.

src/chttpd.erl
kocolosk: derp, thanks

kxepal:
@kocolosk here cannot help you, that's another big story.

kocolosk:
Did a little bit of commit squashing but no changes to the final diff. Going to merge now, thanks all for the review!

PR #39, opened 2015-06-29
Account for binary keys for headers in CORS

iilyak:
We do have headers' keys represented as either lists or binaries.
Make sure we don't crash on `string:to_lower`.

COUCHDB-2733

### DIFF
--- a/src/chttpd_cors.erl
+++ b/src/chttpd_cors.erl
@@ -217,9 +217,13 @@ maybe_apply_headers(CorsHeaders, RequestHeaders) ->
 
 
 simple_headers(Headers) ->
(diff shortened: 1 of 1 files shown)


iilyak:
Fix typo in commit message

rnewson:
+1 after fixing typo "fir" in PR title and commit message.

iilyak:
updated

kxepal:
+1 but could be +10 if we have some tests for this case (:

rnewson:
+1. @iilyak can you merge or do you need someone to do it?

iilyak:
@rnewson: I don't have commit power.

PR #46, opened 2015-07-22
Update mochiweb to fix the problem when second request is ignored

iilyak:
This reverts commit 56cf6412ad3572b5d07a281c5a25ec511af0530c.

Newer version of mochiweb closes connection on failed requests.

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -246,8 +246,6 @@ handle_request(MochiReq0) ->
                 _Else ->
                     send_error(HttpReq, {Error, nil, Stack})
             end
-    after
-        maybe_discard_body(HttpReq)
     end,
 
     RequestTime = timer:now_diff(os:timestamp(), Begin)/1000,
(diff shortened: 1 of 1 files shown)


kxepal:
@iilyak how much newer? Keep in mind, that ours is quite different from upstream one.

iilyak:
> @iilyak how much newer? Keep in mind, that ours is quite different from upstream one.

@kxepal: I was about to issue PR for mochiweb and figure it out. 
We would need this fix https://github.com/mochi/mochiweb/commit/87d6eb15490729698fd192228b8fc963b3777c51

iilyak:
The `maybe_discard_body` approach doesn't work correctly if client (python's requests library) reuses tcp connection.

kxepal:
@rnewson do you remember on what changes we're not in sync with upstream mochiweb? I recall four: timezones, message queue, ssl acceptor, sock5 proxy. Anything else?

May be it would be easy to finally update it...

rnewson:
the socks5 proxy work is for ibrowse not mochiweb iirc, but we can't replace it with the upstream version as it leaks via DNS and thus makes it inappropriate for certain classes of users (Tor users).

kxepal:
Oh, right - that was indeed ibrowser. Then three notable changes so far. Well, actually two: ssl acceptor was fixed in mochiweb half of year ago, but a little different than what we did.

rnewson:
I'm all for getting back to verbatim upstream changes. I never did submit a PR to fix the timezone problem, but we should, it mattered.

iilyak:
@rnewson, @kxepal: Should I close this PR or we are going to use it for discussion.

kxepal:
@iilyak let's keep it until we update mochiweb

iilyak:
> @iilyak let's keep it until we update mochiweb
Does anyone working on it. I don't have the background to decide which customizations worth carrying over.

PR #47, opened 2015-07-22
add configurable blacklist for all URL endpoints

janl:
with help from @iilyak, thanks!

### DIFF
--- a/src/chttpd_handlers.erl
+++ b/src/chttpd_handlers.erl
@@ -30,18 +30,21 @@ provider(App, Module) ->
        App, chttpd_handlers, Module).
 
 url_handler(HandlerKey, DefaultFun) ->
+    ok = ensure_enabled(HandlerKey),
     case collect(url_handler, [HandlerKey]) of
         [HandlerFun] -> HandlerFun;
         [] -> DefaultFun
     end.
 
 db_handler(HandlerKey, DefaultFun) ->
+    ok = ensure_enabled(HandlerKey),
     case collect(db_handler, [HandlerKey]) of
         [HandlerFun] -> HandlerFun;
         [] -> DefaultFun
     end.
 
 design_handler(HandlerKey, DefaultFun) ->
+    ok = ensure_enabled(HandlerKey),
     case collect(design_handler, [HandlerKey]) of
         [HandlerFun] -> HandlerFun;
         [] -> DefaultFun
@@ -50,6 +53,20 @@ design_handler(HandlerKey, DefaultFun) ->
 %% ------------------------------------------------------------------
 %% Internal Function Definitions
 %% ------------------------------------------------------------------
+
+is_disabled(KeyBin) ->
+    BlacklistConfig = config:get("chttpd", "url_handler_blacklist", []),
+    {ok, Blacklist} = couch_util:parse_term(BlacklistConfig),
+    Key = binary_to_list(KeyBin),
+    lists:member(Key, Blacklist).
+
+ ensure_enabled(Key) ->
+    case is_disabled(Key) of
+      true ->
+        throw("The URL endpoint " ++ Key ++ "has been disabled.");
+      false ->
+        ok
+    end.
 
 collect(Func, Args) ->
     Results = do_apply(Func, Args, [ignore_providers]),


kxepal:
How to use it?

src/chttpd_handlers.erl
```diff
@@ -51,6 +54,20 @@ design_handler(HandlerKey, DefaultFun) ->
...
+is_disabled(KeyBin) ->
+    BlacklistConfig = config:get("chttpd", "url_handler_blacklist", []),
+    {ok, Blacklist} = couch_util:parse_term(BlacklistConfig),
+    Key = binary_to_list(KeyBin),
+    lists:member(Key, Blacklist).
+
+ ensure_enabled(Key) ->
+    case is_disabled(Key) of
+      true ->
+        throw("The URL endpoint " ++ Key ++ "has been disabled.");
```
rnewson: `throw({not_found, "The URL` to send a 404?

src/chttpd_handlers.erl
```diff
@@ -51,6 +54,20 @@ design_handler(HandlerKey, DefaultFun) ->
 %% Internal Function Definitions
 %% ------------------------------------------------------------------
 
+is_disabled(KeyBin) ->
+    BlacklistConfig = config:get("chttpd", "url_handler_blacklist", []),
+    {ok, Blacklist} = couch_util:parse_term(BlacklistConfig),
+    Key = binary_to_list(KeyBin),
+    lists:member(Key, Blacklist).
```
rnewson: all a bit expensive per request, how about `case config:get_boolean("chttpd", ?b2l(<<"blacklisted.",KeyBin/binary>>))`?

src/chttpd_handlers.erl
```diff
@@ -51,6 +54,20 @@ design_handler(HandlerKey, DefaultFun) ->
 %% Internal Function Definitions
 %% ------------------------------------------------------------------
 
+is_disabled(KeyBin) ->
+    BlacklistConfig = config:get("chttpd", "url_handler_blacklist", []),
+    {ok, Blacklist} = couch_util:parse_term(BlacklistConfig),
```
kxepal: You'll get badmatch with the default value here:
```
(node1@127.0.0.1)2> BlacklistConfig = config:get("chttpd", "url_handler_blacklist", []).
[]
(node1@127.0.0.1)3> {ok, Blacklist} = couch_util:parse_term(BlacklistConfig).
** exception error: no match of right hand side value {error,{1,erl_parse,["syntax error before: ","'.'"]}}
```

src/chttpd_handlers.erl
kxepal: Like it. Thought, may be "disable.foo = true" or something else in imperative mood.

src/chttpd_handlers.erl
kxepal: +1

iilyak: Please select different value for error so it matches one of the clauses here:
https://github.com/apache/couchdb-chttpd/blob/master/src/chttpd.erl#L715
Possible options are:
- https://github.com/apache/couchdb-chttpd/blob/master/src/chttpd.erl#L729
- https://github.com/apache/couchdb-chttpd/blob/master/src/chttpd.erl#L737

src/chttpd_handlers.erl
```diff
@@ -51,6 +54,20 @@ design_handler(HandlerKey, DefaultFun) ->
 %% Internal Function Definitions
 %% ------------------------------------------------------------------
 
+is_disabled(KeyBin) ->
+    BlacklistConfig = config:get("chttpd", "url_handler_blacklist", []),
```
iilyak: How about dedicating whole section in config for this. Than you would be able to choose error message
Something like:
[chttpd_blacklist]
_users = {not_found, "This endpoint is disabled"}
_global_changes = not_implemented

kxepal: Good idea. But how to avoid collisions when same-named resource is defined on different levels like /_foo and /db/_foo? We don't have such cases, but theoretically this is a limitation.

kxepal: In involve your idea, then we'll end with chttpd_blacklist_global_handlers, chttpd_blacklist_database_handlers, chttpd_blacklist_design_handlers ala httpd defines them now. Not bad decision after all.

rnewson: Hm, I think that's getting complicated. I think we just send 404 for all these things.

src/chttpd_handlers.erl
rnewson: I was aiming for;

 error_info({not_found, Reason}) ->
    {404, <<"not_found">>, Reason};

src/chttpd_handlers.erl
kxepal: Good side of @iilyak idea as it also allows to make previously existed resources [gone](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.11) gracefully. It has a difference from regular 404. Not implemented also may have a point for basement of missed features. But that's all about details.

src/chttpd_handlers.erl
janl: Alternatively, we could cache the parsed list representation somewhere any only access it here. Or is the `lists:member` what you are concerned about?

src/chttpd_handlers.erl
janl: there is a point to be made to differentiate the different levels at which these can be blacklisted to mirror the different levels at which these can be registered:

```
[httpd_design_handlers]
[httpd_global_handlers]
[httpd_db_handlers]
```

Although I’d still prefer to have an actual list, as in:

```
[chttpd]
httpd_design_handlers = []
httpd_global_handlers = []
httpd_db_handlers = []
```

src/chttpd_handlers.erl
janl: A (very) dirty version of this, just to illustrate the idea:

```erlang
get_blacklist() ->
    case config:get("chttpd", "_url_handler_blacklist", []) of
        [] -> % no cached version
            case config:get("chttpd", "url_handler_blacklist", []) of
                [] -> []; % no config value
                BlacklistConfig -> % we have a config value, let’s parse it once
                    {ok, Blacklist} = couch_util:parse_term(BlacklistConfig),
                    config:set("chttpd", "_url_handler_blacklist", Blacklist),
                    Blacklist
            end;
        Blacklist1 -> Blacklist1 % we already have the parsed list cached
    end.

is_disabled(KeyBin) ->
    Blacklist = get_blacklist(),
    Key = ?b2l(KeyBin),
    lists:member(Key, Blacklist).
```

Of course, we’d need a thing that invalidates the “cache”.

I’m abusing the config system here as a cache, we obviously don’t want to do that for real. I’m not too familiar with chttpd yet, would there be an obvious location for this?

rnewson: mochiglobal would be a better spot for a cache, but I think we should avoid additional caching.

src/chttpd_handlers.erl
kxepal: @janl It's not much different from yours:
```
[chttpd_design_handler_blacklist]
foo = not_found

[chttpd_global_handler_blacklist]
bar = not_found

[chttpd_db_handler_blacklist]
baz = not_implemented

```

except that here you can direct match by url key as it is an option name. In your case you'll have to parse and scan list everytime, for hits and misses. 

And reason to have any cache goes away since config structure becomes more friendly.

src/chttpd_handlers.erl
kxepal: I agree with avoid. Cache invalidation is one of the hardest CS problems (: 
Also, you probably mean `config:set("chttpd", "_url_handler_blacklist", Blacklist, false)` - there is no need to have that cache to be persistent.

src/chttpd_handlers.erl
janl: @kxepal I get the benefits there, but I think giving end-users control over what to return *per resource* is taking it too far. We can do this smarter :)

kxepal: @janl If end user decided to black list something, they may (or need in this proposal) define a reason why, right? 

We can do this smarter, but smarter solutions takes more time/resources while that's stupid one just works and quite trivial in implementation. We can always improve it later. Especially, after collecting user experience for this blacklist management - may be they wouldn't touch it a lot (actually, they will) and that decision is good enough.

janl: @kxepal Well, configuration is part of what the end-user sees, so we can’t just fix it later. We need to consider an upgrade plan and everything. I’d like to avoid doing that and I don’t mind spending another day of nailing this one shut so we can forget about it.

But you make a good point, I’ll keep thinking about this.

kxepal: I played a bit with your proposal. One more point against lists: it's hard to operate with them via API. 
You cannot simple remove one endpoint from blacklist: you need to GET `/_config/chttpd/httpd_global_handler_blacklist`, decode, pop element, encode, PUT it back. Same story if you need to add something to blacklist. Compare it with a single PUT/DELETE `/_config/chttpd_global_handlers_blacklist/_uuids` request, which both are atomic.

Simplicity of using that blacklist via API also matters, imho.

janl: I like this. Thanks, I’ll wrap things around.

PR #50, opened 2015-07-29
Don't reset rewrite counter every time

rnewson:
We were missing the handle_request / handle_request_int split that
makes the ported rewrite counter work.


rnewson:
closing in preference to #51

PR #53, opened 2015-07-31
Epi hooks

iilyak:
This is one of the series of PRs to make couchdb plugable. I'll provide complete list of dependent PRs in https://github.com/apache/couchdb-couch repository.

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -480,8 +480,10 @@ qs_json_value(Req, Key, Default) ->
             ?JSON_DECODE(Result)
     end.
 
-qs(#httpd{mochi_req=MochiReq}) ->
-    MochiReq:parse_qs().
+qs(#httpd{mochi_req = MochiReq, qs = undefined}) ->
+    MochiReq:parse_qs();
+qs(#httpd{qs = QS}) ->
+    QS.
 
 path(#httpd{mochi_req=MochiReq}) ->
     MochiReq:get(path).
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -149,8 +149,6 @@ handle_request_int(MochiReq) ->
     % removed, but URL quoting left intact
     RawUri = MochiReq:get(raw_path),
     {"/" ++ Path, _, _} = mochiweb_util:urlsplit_path(RawUri),
-    {HandlerKeyStr, _, _} = mochiweb_util:partition(Path, "/"),
-    HandlerKey = ?l2b(HandlerKeyStr),
 
     Peer = MochiReq:get(peer),
     LogForClosedSocket = io_lib:format("mochiweb_recv_error for ~s - ~p ~s", [
@@ -198,6 +196,12 @@ handle_request_int(MochiReq) ->
         path_parts = [list_to_binary(chttpd:unquote(Part))
                 || Part <- string:tokens(Path, "/")]
     },
+
+    HandlerKey =
+        case HttpReq#httpd.path_parts of
+            [] -> <<>>;
+            [Key|_] -> ?l2b(quote(Key))
+        end,
 
     % put small token on heap to keep requests synced to backend calls
     erlang:put(nonce, couch_util:to_hex(crypto:rand_bytes(5))),
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -193,7 +193,7 @@ handle_request_int(MochiReq) ->
 
     Nonce = couch_util:to_hex(crypto:rand_bytes(5)),
 
-    HttpReq = #httpd{
+    HttpReq0 = #httpd{
         mochi_req = MochiReq,
         begin_ts = Begin,
         peer = Peer,
@@ -203,6 +203,8 @@ handle_request_int(MochiReq) ->
         path_parts = [list_to_binary(chttpd:unquote(Part))
                 || Part <- string:tokens(Path, "/")]
     },
+
+    {ok, HttpReq} = chttpd_plugin:before_request(HttpReq0),
 
     HandlerKey =
         case HttpReq#httpd.path_parts of
@@ -229,7 +231,7 @@ handle_request_int(MochiReq) ->
                     HandlerKey, fun chttpd_db:handle_request/1),
                 AuthorizedReq = chttpd_auth:authorize(possibly_hack(Req),
                     fun chttpd_auth_request:authorize_request/1),
-                HandlerFun(AuthorizedReq);
+                {AuthorizedReq, HandlerFun(AuthorizedReq)};
             Response ->
                 Response
             end;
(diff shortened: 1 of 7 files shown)


src/chttpd.erl
```diff
@@ -724,10 +793,11 @@ error_info({bad_ctype, Reason}) ->
     {415, <<"bad_content_type">>, Reason};
 error_info(requested_range_not_satisfiable) ->
     {416, <<"requested_range_not_satisfiable">>, <<"Requested range not satisfiable">>};
-error_info({error, illegal_database_name}) ->
-    {400, <<"illegal_database_name">>, <<"Only lowercase letters (a-z), "
-        "digits (0-9), and any of the characters _, $, (, ), +, -, and / are "
-        "allowed. Moreover, the database name must begin with a letter.">>};
+error_info({error, {illegal_database_name, Name}}) ->
```
kxepal: I guess if I provide my own `validate_db_name` hook with different regexp constraint, this error message will be true no more, right?

src/chttpd.erl
```diff
@@ -345,28 +405,33 @@ make_uri(Req, Raw) ->
     {[{<<"url">>,Url}, {<<"headers">>,{Headers}}]}.
 %%% end hack
 
+authenticate_request(Req) ->
+    AuthenticationFuns = [
+        {"cookie", fun chttpd_auth:cookie_authentication_handler/1},
+        {"default", fun chttpd_auth:default_authentication_handler/1},
+        {"local", fun chttpd_auth:local_admin_handler/1} %% should be last
```
kxepal: In addition to [these problems](https://github.com/apache/couchdb-couch/pull/81/files#diff-b34b52654741e1d12cd91c2b36b09968L350) you seems eventually drop [`OAuth`](https://github.com/apache/couchdb-couch/blob/master/src/couch_httpd_oauth.erl#L33-L35) support.

kxepal:
Can we use `Resp` or `Response` instead of `Res` when we deal with http responses? Same is `HttpRes` vs `HttpResp`. The problem is that `Res` may reference to "response", "resource", "resolution", but also to a generic "result". One character is not a big cost for making things more easy to read.

src/chttpd_auth.erl
```diff
@@ -12,15 +12,70 @@
...
     couch_httpd_auth:cookie_authentication_handler(Req, chttpd_auth_cache).
 
+local_admin_handler(Req) ->
+    case config:get("chttpd", "require_valid_user", "false") of
+    "true" ->
+        throw({unauthorized, <<"Authentication required.">>});
+    "false" ->
+        case config:get("admins") of
+        [] ->
+            Ctx = #user_ctx{roles=[<<"_reader">>, <<"_writer">>, <<"_admin">>]},
```
kxepal: Shouldn't here be used [?ADMIN_CTX](https://github.com/apache/couchdb-couch/blob/master/include/couch_db.hrl#L41)? Roles list seems doesn't matches, but also a good chance to verify if "_reader" and "_writer" are need while "_admin" is defined.

src/chttpd_auth.erl
```diff
@@ -12,15 +12,70 @@
...
+%% ------------------------------------------------------------------
+
 default_authentication_handler(Req) ->
     couch_httpd_auth:default_authentication_handler(Req, chttpd_auth_cache).
 
 cookie_authentication_handler(Req) ->
     couch_httpd_auth:cookie_authentication_handler(Req, chttpd_auth_cache).
 
+local_admin_handler(Req) ->
+    case config:get("chttpd", "require_valid_user", "false") of
```
kxepal: So, since you touched this line, bikeshedding question: we have same option defined in "couch_http_auth" [\[1\]](https://github.com/apache/couchdb-couch/blob/master/src/couch_httpd_auth.erl#L106) [\[2\]](https://github.com/apache/couchdb-couch/blob/master/src/couch_httpd.erl#L409)  section. Shouldn't it be used here as well to avoid having same things duplicated?

iilyak: Not sure about this one. Since backdoor interface in production is behind a firewall. I can imagine setups where for backdoor interface `require_valid_user` is `false` while it is `true` for clustered.

kxepal: Ok, let's move discussion about to another PR/issue/whatever. I raised this question to not forget about it in future.

src/chttpd.erl
```diff
@@ -195,30 +191,46 @@ handle_request_int(MochiReq) ->
...
+        original_method = Method1,
+        nonce = Nonce,
         method = Method,
         path_parts = [list_to_binary(chttpd:unquote(Part))
                 || Part <- string:tokens(Path, "/")]
     },
 
+    {ok, HttpReq} = chttpd_plugin:before_request(HttpReq0),
+
+    HandlerKey =
```
rnewson: this change appears unrelated to the PR at hand. I understand the desire to clean up code but mixing in unrelated work makes reviews harder.

src/chttpd.erl
```diff
@@ -724,10 +794,13 @@ error_info({bad_ctype, Reason}) ->
     {415, <<"bad_content_type">>, Reason};
 error_info(requested_range_not_satisfiable) ->
     {416, <<"requested_range_not_satisfiable">>, <<"Requested range not satisfiable">>};
-error_info({error, illegal_database_name}) ->
-    {400, <<"illegal_database_name">>, <<"Only lowercase letters (a-z), "
-        "digits (0-9), and any of the characters _, $, (, ), +, -, and / are "
-        "allowed. Moreover, the database name must begin with a letter.">>};
+error_info({error, {illegal_database_name, Name}}) when is_binary(Name) ->
+    error_info({error, {illegal_database_name, ?b2l(Name)}});
+error_info({error, {illegal_database_name, Name}}) ->
```
rnewson: Name is always binary, according to the epi PR at https://github.com/apache/couchdb-couch/pull/81, so let's drop the list clause and construct this as a binary.

src/chttpd_auth.erl
```diff
@@ -12,15 +12,69 @@
...
+%% Default callbacks
+%% ------------------------------------------------------------------
+
 default_authentication_handler(Req) ->
     couch_httpd_auth:default_authentication_handler(Req, chttpd_auth_cache).
 
 cookie_authentication_handler(Req) ->
     couch_httpd_auth:cookie_authentication_handler(Req, chttpd_auth_cache).
 
+local_admin_handler(Req) ->
```
rnewson: `admin_handler` here?

rnewson: actually, this should be called `party_mode_handler`, it exists only to grant admin rights to any user if there are no admins, it does not authenticate admins.

test/chttpd_error_info_tests.erl
```diff
@@ -110,11 +110,11 @@ error_info_test() ->
              <<"Requested range not satisfiable">>}
         },
         {
-            {error, illegal_database_name},
+            {error, {illegal_database_name, "foo"}},
```
rnewson: send a binary, since that's what happens for real.

rnewson:
I'm generally +1, great work. I made a few points that I'd like to see addressed before merging. And a final note about keeping unrelated work out of PR's in future.

src/chttpd.erl
iilyak: This change is related to PR. Previously HandlerKey was constructed using `mochiweb_util:partition(Path, "/")`. 
I had to change the logic to use first part from `path_parts` field of `#httpd{}` record because `#httpd{}` record doesn't have field for full path. I couldn't just use old way either since plugin might want to manipulate Path in order to affect handler selection. Therefore HandlerKey calculation has to happen after `chttpd_plugin:before_request` and it should be based on the value returned from that call.

rnewson:
+1

PR #56, opened 2015-08-14
check POST requests for valid json header

mayya-sharipova:
validate that all POST requests with json body must have also have valid
json header: {"Content-Type": "application/json"}
This ensures a basic protection against CSRF

JIRA: COUCHDB-2775

### DIFF
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -73,7 +73,7 @@ handle_request(#httpd{path_parts=[DbName|RestParts],method=Method}=Req)->
     end.
 
 handle_changes_req(#httpd{method='POST'}=Req, Db) ->
-    couch_httpd:validate_ctype(Req, "application/json"),
+    chttpd:validate_ctype(Req, "application/json"),
     handle_changes_req1(Req, Db);
 handle_changes_req(#httpd{method='GET'}=Req, Db) ->
     handle_changes_req1(Req, Db);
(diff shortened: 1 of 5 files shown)


src/chttpd_db.erl
```diff
@@ -362,7 +362,6 @@ db_req(#httpd{path_parts=[_,<<"_ensure_full_commit">>]}=Req, _Db) ->
 
 db_req(#httpd{method='POST',path_parts=[_,<<"_bulk_docs">>], user_ctx=Ctx}=Req, Db) ->
     couch_stats:increment_counter([couchdb, httpd, bulk_requests]),
-    couch_httpd:validate_ctype(Req, "application/json"),
```
kxepal: Why did you remove the check here?

mayya-sharipova: Because after this line follows:  {JsonProps} = chttpd:json_body_obj(Req), and the function json_body_obj already having a check for header.

kxepal: mmm...I think it's conceptually wrong: `chttpd:json_body_obj` ensures that request body is a JSON object, but there is more general `json_body` function which need to be insured that request body _is_ a JSON value. 

However, both shouldn't care about mime type - that's not their work and this prevents to reuse these function if request body is a JSON, but mime type is not a `application/json`. For instance, it could be `application/vnd.cloudant+json`, but because of `chttpd:json_body` does additional wrong job, you'll have to create your own function to decode JSON. 

Single responsibility is a good idea to follow.

mayya-sharipova: Thanks for the detailed comments, this makes sense. I have updated the commit to reflect them.

kxepal:
LGFM. Did you check other apps which provides chttpd endpoints for POST content-type validation?

mayya-sharipova:
Thanks for the comments. Not sure what LGFM means. For the other apps: I plan to validate headers for POST requests also in couch, couch_mrview and mango.

kxepal:
@mayya-sharipova Looks Good For Me (:
@rnewson +1?

rnewson:
I'd rather this was chttpd:validate_ctype instead, but it looks good otherwise. Let's add validate_ctype to chttpd and have that function, and only that one, call down to couch_httpd.

rnewson:
+1

PR #59, opened 2015-08-20
handle_system_req Memory includes total

jcounts:
@rnewson 

Originally was done here: https://github.com/cloudant/chttpd

moved to this repo

### DIFF
--- a/src/chttpd_misc.erl
+++ b/src/chttpd_misc.erl
@@ -329,7 +329,7 @@ flush(Node, Req) ->
 % because it's not couch trunk
 handle_system_req(Req) ->
     Other = erlang:memory(system) - lists:sum([X || {_,X} <-
-        erlang:memory([atom, code, binary, ets])]),
+        erlang:memory([atom, code, binary, ets, total])]),
     Memory = [{other, Other} | erlang:memory([atom, atom_used, processes,
         processes_used, binary, code, ets])],
     {NumberOfGCs, WordsReclaimed, _} = statistics(garbage_collection),


kxepal:
How can I trigger this handler via HTTP API? So far I see there is no place in code that makes it available to handle incoming requests.

rnewson:
@kxepal: does seem to be missing from https://github.com/apache/couchdb-chttpd/blob/master/src/chttpd_httpd_handlers.erl :/

kxepal:
@rnewson Yes, I noticed that, but commentary for the handler says that it's only available for backdoor interface...or at least it should be.

I'll add it to default.ini where global httpd handlers are defined and ensure what it's available by `/_node/X/_system` path on cluster interface, ok?

rnewson:
ah, quite right. Let's just add it to :5986 for now

rnewson:
actually /_system is available on both front and back ports at Cloudant, I just checked, so we should do the same.

kxepal:
@rnewson ok, I'll made PRs for this.

kxepal:
Hm, not sure what's use case for this patch. Example output:
```
    "memory": {
        "atom": 801697,
        "atom_used": 793412,
        "binary": 1656328,
        "code": 23299745,
        "ets": 1753608,
        "other": -36439634,
        "processes": 8933048,
        "processes_used": 8919640
    }
```
Notice negative other value. I believe, total value should be shows with separate object field.

jcounts:
hmm...I wasn't sure if this would solve my problem or not. This was a response for:
https://cloudant.fogbugz.com/f/cases/47888/DBcore-memory-usage-displayed-in-admin-dashboard-is-wrong

Which is a bug in:
https://github.com/cloudant/wilson/blob/develop/wilson/views/clusters.py

The Wilson endpoint pulls info from graphite (and haproxy).

kxepal:
@jcounts Sorry, I don't have access to Cloudant fogbugz and private repositories. What's the case there is short which is fine to gather negative memory values?

kxepal:
@jcounts also, how about to backport those fogbugz ticket to CouchDB JIRA and bound it with this commit? Would be helpful to track the things.

PR #60, opened 2015-08-21
Ensure that provided bind_address is correct

kxepal:
`invalid_bind_address` error is more helpful than `badmatch` from deeps of mochiweb internals.

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -109,7 +109,11 @@ start_link(https) ->
     start_link(https, Options).
 
 start_link(Name, Options) ->
-    IP = with_default(config:get("chttpd", "bind_address"), any),
+    IP = case config:get("chttpd", "bind_address", "any") of
+             "any" -> any;
+             Else -> Else
+         end,
+    ok = couch_httpd:validate_bind_address(IP),
 
     Options1 = Options ++ [
         {loop, fun ?MODULE:handle_request/1},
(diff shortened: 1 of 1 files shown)


rnewson:
+1

src/chttpd.erl
```diff
@@ -109,7 +109,8 @@ start_link(https) ->
     start_link(https, Options).
 
 start_link(Name, Options) ->
-    IP = with_default(config:get("chttpd", "bind_address"), any),
+    IP = config:get("chttpd", "bind_address", "any"),
+    couch_httpd:validate_bind_address(IP),
```
rnewson: you need to use the return value of validate_bind_address here (to have it converted to atom any)

kxepal: AAA! Indeed I miss that moment. Thanks.

src/chttpd.erl
```diff
@@ -109,7 +109,11 @@ start_link(https) ->
     start_link(https, Options).
 
 start_link(Name, Options) ->
-    IP = with_default(config:get("chttpd", "bind_address"), any),
+    IP = case config:get("chttpd", "bind_address", "any") of
+             "any" -> any;
+             Else -> Else
+         end,
+    couch_httpd:validate_bind_address(IP),
```
rnewson: say `ok =` here

rnewson:
+1

PR #61, opened 2015-08-21
Expose /_system handler

kxepal:


src/chttpd_httpd_handlers.erl
```diff
@@ -25,6 +25,7 @@ url_handler(<<"_replicate">>)      -> fun chttpd_misc:handle_replicate_req/1;
 url_handler(<<"_uuids">>)          -> fun chttpd_misc:handle_uuids_req/1;
 url_handler(<<"_session">>)        -> fun chttpd_auth:handle_session_req/1;
 url_handler(<<"_up">>)             -> fun chttpd_misc:handle_up_req/1;
+url_handler(<<"_system">>)         -> fun chttpd_misc:handle_system_req/1;
```
rnewson: we don't want this here, we want a 404.

rnewson:
+1 after removing /_system. We want that to 404, the correct method is /_node/<fqdn>/_system as this works on any node for any node.

src/chttpd_httpd_handlers.erl
kxepal: ok, fixed

rnewson:
+1. I agree we should move this handle to couch/src/couch_httpd_ instead, but not blocking the release.

kxepal:
Ok, will move it later with next iteration.

kxepal:
Closed with 936e8ee

PR #62, opened 2015-08-21
add clusterwide compaction for dbs

robertkowalski:
COUCHDB-2492

Still missing: per view compaction

### DIFF
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -225,9 +225,20 @@ maybe_flush_changes_feed(Acc0, Data, Len) ->
     },
     {ok, Acc}.
 
-handle_compact_req(Req, _) ->
-    Msg = <<"Compaction must be triggered on a per-shard basis in CouchDB">>,
-    couch_httpd:send_error(Req, 403, forbidden, Msg).
+handle_compact_req(#httpd{method='POST', path_parts=[_DbName, <<"_compact">>, _DesignName | _]}=Req, _Db) ->
+    Msg = <<"Compactions for views must be triggered on a per-shard basis in CouchDB">>,
+    couch_httpd:send_error(Req, 403, forbidden, Msg);
+
+handle_compact_req(#httpd{method='POST', path_parts=[_DbName, <<"_compact">>]}=Req, Db) ->
+    couch_httpd:validate_ctype(Req, "application/json"),
+    ok = couch_db:check_is_admin(Db),
+    couch_httpd:body(Req),
+    fabric:compact_db(Db#db.name),
+    send_json(Req, 202, {[{ok, true}]});
+
+handle_compact_req(Req, _Db) ->
+    send_method_not_allowed(Req, "POST").
+
 
 handle_view_cleanup_req(Req, Db) ->
     ok = fabric:cleanup_index_files(Db),


src/chttpd_db.erl
```diff
@@ -225,9 +225,19 @@ maybe_flush_changes_feed(Acc0, Data, Len) ->
...
 
-handle_compact_req(Req, _) ->
-    Msg = <<"Compaction must be triggered on a per-shard basis in CouchDB">>,
-    couch_httpd:send_error(Req, 403, forbidden, Msg).
+handle_compact_req(#httpd{method='POST'}=Req, Db) ->
+    case Req#httpd.path_parts of
+        [_DbName, <<"_compact">>] ->
+            ok = couch_db:check_is_admin(Db),
+            couch_httpd:validate_ctype(Req, "application/json"),
+            _ = couch_httpd:body(Req),
```
eiri: If you are not interested in the response, why do you need an assign to pass through? :)

kxepal: I believe it's intentional. See https://github.com/apache/couchdb-couch/pull/36 for the case.

src/chttpd_db.erl
```diff
@@ -225,9 +225,19 @@ maybe_flush_changes_feed(Acc0, Data, Len) ->
     },
     {ok, Acc}.
 
-handle_compact_req(Req, _) ->
-    Msg = <<"Compaction must be triggered on a per-shard basis in CouchDB">>,
-    couch_httpd:send_error(Req, 403, forbidden, Msg).
+handle_compact_req(#httpd{method='POST'}=Req, Db) ->
+    case Req#httpd.path_parts of
```
kxepal: Why not to match path_parts with the function signature?

kxepal:
Does it works?
```
Compiled src/chttpd_httpd_handlers.erl
/home/kxepal/projects/asf/couchdb/couchdb/src/chttpd/src/chttpd_db.erl:236: syntax error before: 'end'
/home/kxepal/projects/asf/couchdb/couchdb/src/chttpd/src/chttpd_db.erl:17: function handle_compact_req/2 undefined
Compiling /home/kxepal/projects/asf/couchdb/couchdb/src/chttpd/src/chttpd_db.erl failed:
```

src/chttpd_db.erl
eiri: I understand what call to `body` does, but assign to untagged match all has at most a decorative value.

src/chttpd_db.erl
robertkowalski: will add another case for view compaction soon

robertkowalski:
> Does it works?

ah good catch, had some leftovers from the changes i already made for view copaction

src/chttpd_db.erl
```diff
@@ -225,9 +225,23 @@ maybe_flush_changes_feed(Acc0, Data, Len) ->
     },
     {ok, Acc}.
 
-handle_compact_req(Req, _) ->
-    Msg = <<"Compaction must be triggered on a per-shard basis in CouchDB">>,
-    couch_httpd:send_error(Req, 403, forbidden, Msg).
+handle_compact_req(#httpd{method='POST'}=Req, Db) ->
+    couch_httpd:validate_ctype(Req, "application/json"),
+    case Req#httpd.path_parts of
```
rnewson: this is the first time I've seen path matching within a function, everywhere else we do this at the top level of the function. So, please make this two separate clauses.

robertkowalski: interesting, i took the original compact handler at https://github.com/apache/couchdb-couch/blob/aa5d63fce03f6bac18e7c1ffbbaf84f898baf010/src/couch_httpd_db.erl#L161 as an inspiration. will change it

src/chttpd_db.erl
```diff
@@ -225,9 +225,24 @@ maybe_flush_changes_feed(Acc0, Data, Len) ->
     },
     {ok, Acc}.
 
-handle_compact_req(Req, _) ->
-    Msg = <<"Compaction must be triggered on a per-shard basis in CouchDB">>,
-    couch_httpd:send_error(Req, 403, forbidden, Msg).
+handle_compact_req(#httpd{
+        method='POST', path_parts=[_DbName, <<"_compact">>, _DesignName | _]
```
rnewson: fix indentation. best to have this on one line even if's long.

rnewson:
+1 after indentation and squash

rnewson:
changing my mind on this to -1, I commented on the jira ticket COUCHDB-2492 with more justification.

robertkowalski:
feel free to revert it

robertkowalski:
its on master already

robertkowalski:
added some comments to the ticket

PR #65, opened 2015-08-25
Refactor test suite for httpd endpoints

iilyak:
Split out common parts into util module so we can reuse test suite.
The util module could be used for testing:
- endpoints defined in other applications
- overridden endpoints

COUCHDB-2789


iilyak:
This PR depends on https://github.com/apache/couchdb-couch-epi/pull/6

src/chttpd_httpd_handlers_test_util.erl
```diff
@@ -0,0 +1,200 @@
...
+%%%=========================================================================
+%%%  Test environment defintions
+%%%=========================================================================
+
+start(App, Apps) ->
+    Ctx = test_util:start_couch(Apps),
+    wait_handlers(App),
+    Ctx.
+
+setup("mocked") ->
```
kxepal: Why not use atoms here?

test/chttpd_httpd_handlers_tests.erl
```diff
@@ -0,0 +1,54 @@
...
+        {<<"_utils">>, chttpd_misc, handle_utils_dir_req},
+        {<<"_all_dbs">>, chttpd_misc, handle_all_dbs_req},
+        {<<"_active_tasks">>, chttpd_misc, handle_task_status_req},
+        {<<"_node">>, chttpd_misc, handle_node_req},
+        {<<"_reload_query_servers">>, chttpd_misc, handle_reload_query_servers_req},
+        {<<"_replicate">>, chttpd_misc, handle_replicate_req},
+        {<<"_uuids">>, chttpd_misc, handle_uuids_req},
+        {<<"_session">>, chttpd_auth, handle_session_req},
+        {<<"_up">>, chttpd_misc, handle_up_req},
+        {<<"anything">>, chttpd_db, handle_request}
```
kxepal: I wonder if there is a way to not have this list duplicate definitions in chttpd_chttpd_handlers...at least for handlers.

iilyak: It would be trivial with elixir macro :-).

On a serious note though few ideas to consider (crazy ones as well).
1. Use https://github.com/richcarl/merl to generate `handlers` clauses
2. Extract the info from beam chunks somehow
3. Use macro somehow (very problematic with erlang text based preprocessor).
4. Define handlers as it is done in tests (performance impact).
5. Require something like `endpoints(db_handler) -> [<<"_view_cleanup">>, <<"compact">>]` in `<app>_httpd_handlers.erl`

iilyak: 5. Is the simplest one

iilyak: On a side note. I am also thinking rather it would be better to place tests in `<app>_httpd_handlers.erl` itself.

iilyak: I'm actually quite like option 5)

kxepal: > I am also thinking rather it would be better to place tests in <app>_httpd_handlers.erl itself.

+1 - this will at least help to not miss the tests if you're going to modify handlers. My point here was about to not fall into situation, when you change/update `<app>_httpd_handlers.erl` and not update the tests (they may eventually stay green after) and imho this should help a bit.

> It would be trivial with elixir macro :-).

I think use elixir for testing is not a bad idea at all. They have quite friendly and powerful tools for that. But not today (:

iilyak:
I did rework the structure of the PR a little bit. 
Now it has mandatory `endpoint/1` function in every `<app>_httpd_handlers.erl` to facilitate introspection.
Let me know what you think.

kxepal:
I think it's good and helpful for introspection proposes. And one more step closer to the same behavior as `httpd_*_handlers` config options provides now (:

janl:
@iilyak @davisp are you planning to bring this up to master, or can we close this?

iilyak:
stale PR

PR #67, opened 2015-09-02
2796 improve performance of providers function

iilyak:
This PR depends on apache/couchdb-couch-epi#8


rnewson:
+1 after main PR is +1'ed.

PR #69, opened 2015-09-08
Dont insert nil

rnewson:


kxepal:
LGFM. Would you like to handle other tabs in the same module with "remove tabs" commit?

rnewson:
I think for today I just want to fix the tabs in that function, as they made the functional change diff look very weird. We should do a full sweep and convert all tabs at some point.

PR #72, opened 2015-09-11
Allow filtering of active tasks by property

rnewson:
COUCHDB-2807

### DIFF
--- a/src/chttpd_misc.erl
+++ b/src/chttpd_misc.erl
@@ -130,6 +130,13 @@ all_dbs_callback({error, Reason}, #vacc{resp=Resp0}=Acc) ->
     {ok, Resp1} = chttpd:send_delayed_error(Resp0, Reason),
     {ok, Acc#vacc{resp=Resp1}}.
 
+handle_task_status_req(#httpd{method='GET', path_parts=[_, K, V]}=Req) ->
+    {Replies, _BadNodes} = gen_server:multi_call(couch_task_status,
+        {filter, K, V}),
+    Response = lists:flatmap(fun({Node, Tasks}) ->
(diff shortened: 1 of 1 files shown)


kxepal:
I think API is incorrect since you use HTTP resources as filters, not to define some new thing. I think filtering should be done with query parameters, which will allow to filter by multiple properties, like "show me all replications for node1 which run by user foo against target bar".

kxepal:
Also, it seems not all properties could be used for filtering:
```
$ http --json http://localhost:15984/_active_tasks/continuous/true
[
    {
        "changes_pending": 1692895,
        "checkpoint_interval": 5000,
        "checkpointed_source_seq": 57182,
        "continuous": true,
        "doc_id": null,
        "doc_write_failures": 0,
        "docs_read": 6118,
        "docs_written": 6118,
        "missing_revisions_found": 6118,
        "node": "node1@127.0.0.1",
        "pid": "<0.1641.0>",
        "replication_id": "a2d6fd6b275527acdf245abca30d7597+continuous+create_target",
        "revisions_checked": 6118,
        "source": "http://localhost:5984/hashes/",
        "source_seq": 57182,
        "started_on": 1442074677,
        "target": "http://127.0.0.1:15984/hashes/",
        "through_seq": 57182,
        "type": "replication",
        "updated_on": 1442075004,
        "user": null
    }
]

$ http --json http://localhost:15984/_active_tasks/node/node1%40127.0.0.1
[]

$ http --json http --json http://localhost:15984/_active_tasks/pid/%3C0.1641.0%3E
[]
```

rnewson:
I don't like it after all

PR #73, opened 2015-09-16
POST to /{db}/_all_docs with invalid keys should return 400

nickva:
Now they return 500. Since there is nothing wrong on the server,
400 is more appropriate.

For example, turns requests like these:

http -a adm:pass POST http://127.0.0.1:15984/db1/_all_docs  keys:='["1",2]'
HTTP/1.1 500 Internal Server Error

{
    "error": "2",
    "reason": "{illegal_docid,2}"
}

Into:

http -a adm:pass POST http://127.0.0.1:15984/db1/_all_docs  keys:='["1",2]'
HTTP/1.1 400 Bad Request

{
    "error": "illegal_docid",
    "reason": 2
}

COUCHDB-2815

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -818,6 +818,8 @@ error_info({error, {illegal_database_name, Name}}) ->
         "digits (0-9), and any of the characters _, $, (, ), +, -, and / ",
         "are allowed. Must begin with a letter.">>,
     {400, <<"illegal_database_name">>, Message};
+error_info({_DocID,{illegal_docid,DocID}}) ->
+    {400, <<"illegal_docid">>,DocID};
 error_info({missing_stub, Reason}) ->
     {412, <<"missing_stub">>, Reason};
 error_info(request_entity_too_large) ->
--- a/test/chttpd_error_info_tests.erl
+++ b/test/chttpd_error_info_tests.erl
@@ -116,6 +116,10 @@ error_info_test() ->
                " the characters _, $, (, ), +, -, and / are allowed."
                " Must begin with a letter.">>}
         },
+	{
+            {Error, {illegal_docid,1}},
+            {400, <<"illegal_docid">>, 1}
+	},
         {
             {missing_stub, Reason},
             {412, <<"missing_stub">>, Reason}


kxepal:
Hm, it seems we have broken compatibility here. What CouchDB 1.6.1 respond on the same request:

```
$ http POST http://127.0.0.1:5984/db/_all_docs keys:='["1",2]'
HTTP/1.1 200 OK
Date: Wed, 16 Sep 2015 21:08:23 GMT
ETag: "2EDF75VE1NJ20YC70YR128RDV"
Server: CouchDB/1.6.1 (Erlang OTP/18)
Transfer-Encoding: chunked

{
    "offset": 0,
    "rows": [
        {
            "id": "1",
            "key": "1",
            "value": {
                "deleted": true,
                "rev": "4-8bce95a01cc91f50756eb75737dd3946"
            }
        },
        {
            "error": "not_found",
            "key": 2
        }
    ],
    "total_rows": 26
}
```

I think 2.0 should follow the same algorithm which is more friendly for eventual user mistakes.

kxepal:
However, your changes here are also important since invalid document id shouldn't cause server error as it's user one in 99.9% of cases. So I'm +1 here, but /_all_docs still have to be fixed additionally.

nickva:
Kxepal, I see your point. I thought of the compatibility issue as well, and saw the behavior when tested it under 1.x version with Cloudant's code. But a 400 result for whole request seems more right, as it lets the user know that it is not that the key is simply not there, but the request is malformed.

rnewson:
I'd prefer to see `{ "error": "illegal_docid", "reason": 2 }` where couchdb sends `{"error": "not_found",       "key": 2 }` in the response, that seems to be all that's needed here.

in 3.0 I really want to fix up all our request/response handling by using a framework that enforces the same semantics for all endpoints, at that point we can make many new, helpful 400 responses for bad requests.

nickva:
As another point of reference, I compared behavior with _bulk_docs on 2.0 and it seems to return 400 if at least one key is malformed (which would be consistent with this pr). 

```
$ http -a adm:pass POST http://127.0.0.1:15984/db1/_bulk_docs docs:='[ { "_id":3 }, { "_id":"4" } ]'
HTTP/1.1 400 Bad Request
Content-Length: 64
Date: Thu, 17 Sep 2015 14:36:19 GMT
Server: CouchDB/b8b9968 (Erlang OTP/17)
X-Couch-Request-ID: 2fd36db438
X-CouchDB-Body-Time: 0

{
    "error": "bad_request",
    "reason": "Document id must be a string"
}

nvatama@nvm chttpd (master) $ http -a adm:pass POST http://127.0.0.1:15984/db1/_bulk_docs docs:='[ { "_id":"3" }, { "_id":"4" } ]'
HTTP/1.1 201 Created
Content-Length: 130
Date: Thu, 17 Sep 2015 14:36:56 GMT
Server: CouchDB/b8b9968 (Erlang OTP/17)
X-Couch-Request-ID: 9adef1b5e8
X-CouchDB-Body-Time: 0

[
    {
        "id": "3",
        "ok": true,
        "rev": "1-967a00dff5e02add41819138abb3284d"
    },
    {
        "id": "4",
        "ok": true,
        "rev": "1-967a00dff5e02add41819138abb3284d"
    }
]
```

rnewson:
2.0 does that because we pulled that up from dbcore. The topic here is what _should_ we do.

rnewson:
+1 after #couchdb-dev discussion

kxepal:
+1

PR #74, opened 2015-09-17
Use illegal_docid error for really bad document ids

kxepal:


kxepal:
See also: https://github.com/apache/couchdb-couch/pull/101

tonysun83:
just curious why we're adding an extra match here in addition to this change:

https://github.com/apache/couchdb-chttpd/pull/73/files

kxepal:
@tonysun83 

First match `{illegal_docid, Reason}` is for direct `couch_doc:validate_docid/1` call.
Second match goes for all_docs view and caused by some fabric internals. I didn't found where and why it's so awkward.

tonysun83:
ah, from the other PR you referenced above. I see it now.
+1.

rnewson:
+1

PR #75, opened 2015-09-21
Reject database names with path segments over 128 characters

rnewson:
COUCHDB-2821

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -820,6 +820,9 @@ error_info({error, {illegal_database_name, Name}}) ->
     {400, <<"illegal_database_name">>, Message};
 error_info({_DocID,{illegal_docid,DocID}}) ->
     {400, <<"illegal_docid">>,DocID};
+error_info({error, {database_name_too_long, DbName}}) ->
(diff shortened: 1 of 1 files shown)


src/chttpd.erl
```diff
@@ -820,6 +820,10 @@ error_info({error, {illegal_database_name, Name}}) ->
     {400, <<"illegal_database_name">>, Message};
 error_info({_DocID,{illegal_docid,DocID}}) ->
     {400, <<"illegal_docid">>,DocID};
+error_info({error, {database_name_too_long, DbName}}) ->
+    {400, database_name_too_long,
```
eiri: Is the second tuple's element is intentionally an atom? It seems to be a binary at the rest of the responses.

kxepal: Suddenly, atoms here works as expected. Not sure why we use binaries for else.

kxepal:
+1

kxepal:
(there is something really wrong with couch_log)

src/chttpd.erl
rnewson: I can change to a binary, old habits.

PR #77, opened 2015-09-28
Update to new couch_epi API

iilyak:
Depends on https://github.com/apache/couchdb-couch-epi/pull/12

### DIFF
--- a/src/chttpd_sup.erl
+++ b/src/chttpd_sup.erl
@@ -18,21 +18,23 @@
 
 %% Helper macro for declaring children of supervisor
 -define(CHILD(I, Type), {I, {I, start_link, []}, permanent, 100, Type, [I]}).
--define(NAMED_CHILD(I, M, Type), {I, {M, start_link, []}, permanent, 100, Type, [M]}).
 
 start_link(Args) ->
     supervisor:start_link({local,?MODULE}, ?MODULE, Args).
 
 init([]) ->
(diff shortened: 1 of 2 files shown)


src/chttpd_sup.erl
```diff
@@ -57,10 +54,8 @@ lru_opts() ->
...
     end.
 
-auth_cache_handler() ->
-    case application:get_env(chttpd, auth_cache) of
-        {ok, Module} ->
-            Module;
-        _ ->
-            chttpd_auth_cache
-    end.
+add_auth_cache_if_missing(Children) ->
```
rnewson: this is in case `couch_epi:register_service(chttpd_epi)` returns a replacement but it's ad-hoc. Would it a better api to pass the pre-epi children in and then use the return value instead? that way chttpd_epi plugins could remove unwanted children cleanly. something like;

```
Children0 = [?CHILD(auth_cache_handler(), worker), and others...],
Children1 = couch_epi:register_service(chttpd_epi, Children0),
{ok, {{one_for_one, 3, 10}, Children1}.
```
register_service would modify the input list, replacing the auth_cache childspec.

iilyak: Implemented in https://github.com/apache/couchdb-couch-epi/commit/33e8bae844e0994f024ee067196161d61872fcfb

PR #78, opened 2015-10-01
Remove chttpd_handlers:provider

iilyak:
`provider` function is no longer needed since providers defined by
couch_epi_plugin behaviour.


kxepal:
+1

PR #82, opened 2015-10-14
COUCHDB-769: Store attachments in the external storage.

gilv:
Initial implementation that allows CouchDB to store attachments outside of the database file.
This implementation supports OpenStack Swift and SoftLayer Object store

### DIFF
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -407,7 +407,15 @@ db_req(#httpd{method='POST',path_parts=[_,<<"_bulk_docs">>], user_ctx=Ctx}=Req, 
         true  -> [all_or_nothing|Options];
         _ -> Options
         end,
-        case fabric:update_docs(Db, Docs, Options2) of
+        couch_log:debug("chttpd_db: update_docs wellcome. Going to store attachment in external store",[]),
+        case fabric_attachments_handler:externalize_att(Db) of
+            "true" ->
+                NewDocs = fabric_attachments_handler:inline_att_store(Db,fabric:docs(Docs));
+            "false" ->
+                couch_log:debug("Store inline attachmets in object store disabled",[]),
+                NewDocs = Docs
+        end,
+        case fabric:update_docs(Db, NewDocs, Options2) of
         {ok, Results} ->
             % output the results
             DocResults = lists:zipwith(fun update_doc_result_to_json/2,
@@ -733,7 +741,17 @@ db_doc_req(#httpd{method='POST', user_ctx=Ctx}=Req, Db, DocId) ->
     NewDoc = Doc#doc{
         atts = UpdatedAtts ++ OldAtts2
     },
-    case fabric:update_doc(Db, NewDoc, Options) of
+    couch_log:debug("chttpd_db: couchNew doc formed including new att: formed",[]),
+    lists:foreach(fun(X) -> couch_log:debug("~p~n ~p~n",[couch_att:fetch(data, X), couch_att:fetch(name, X)]) end, NewDoc#doc.atts),
+    couch_log:debug("chttpd_db: going to call attachment handler for base64 attachments",[]),
+    case fabric_attachments_handler:externalize_att(Db) of
+        "true" ->
+            NewNewDoc = fabric_attachments_handler:inline_att_store(Db, NewDoc); %store_single_document
+		"false" ->
+            couch_log:debug("Store inline attachmets in Swift disabled",[]),
+            NewNewDoc = NewDoc
+    end,
+    case fabric:update_doc(Db, NewNewDoc, Options) of
     {ok, NewRev} ->
         HttpCode = 201;
     {accepted, NewRev} ->
@@ -1079,7 +1097,17 @@ db_attachment_req(#httpd{method='GET',mochi_req=MochiReq}=Req, Db, DocId, FileNa
     case [A || A <- Atts, couch_att:fetch(name, A) == FileName] of
     [] ->
         throw({not_found, "Document is missing attachment"});
-    [Att] ->
+    [TmpAtt] ->
+        AttExternal = couch_att:fetch(att_external,TmpAtt),
+        couch_log:debug("chttpd_db: attachment stored in external store ~p~n",[AttExternal]),
+        case AttExternal of
+            "external" ->
+                Att = fabric_attachments_handler:inline_att_handler(get, Db, TmpAtt),
+                couch_log:debug("chtttpd_db: Got attachment from external store",[]);
+            _ ->
+                couch_log:debug("chttpd_db: Att is not stored in external store",[]),
+                Att = TmpAtt
+        end,
         [Type, Enc, DiskLen, AttLen, Md5] = couch_att:fetch([type, encoding, disk_len, att_len, md5], Att),
         Refs = monitor_attachments(Att),
         try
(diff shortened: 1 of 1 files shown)


src/chttpd_db.erl
```diff
@@ -407,7 +407,15 @@ db_req(#httpd{method='POST',path_parts=[_,<<"_bulk_docs">>], user_ctx=Ctx}=Req,
         true  -> [all_or_nothing|Options];
         _ -> Options
         end,
-        case fabric:update_docs(Db, Docs, Options2) of
+        couch_log:debug("chttpd_db: update_docs wellcome. Going to store attachment in external store",[]),
```
kxepal: welcome! (:

kxepal:
Please fix:
- Style: try to fit 80 chars line length, put space after comma, no tabs, correctly aligned clauses etc.
- Reduce amount of debug logs: debug logs should help you understand what going on in your code and in which state it is in places where problems may happens, not produce noise about every logical step.

kxepal:
Would be nice to see some tests for this to prove that this all works.

src/chttpd_db.erl
```diff
@@ -733,7 +741,17 @@ db_doc_req(#httpd{method='POST', user_ctx=Ctx}=Req, Db, DocId) ->
...
-    case fabric:update_doc(Db, NewDoc, Options) of
+    couch_log:debug("chttpd_db: couchNew doc formed including new att: formed",[]),
+    lists:foreach(fun(X) -> couch_log:debug("~p~n ~p~n",[couch_att:fetch(data, X), couch_att:fetch(name, X)]) end, NewDoc#doc.atts),
+    couch_log:debug("chttpd_db: going to call attachment handler for base64 attachments",[]),
+    case fabric_attachments_handler:externalize_att(Db) of
+        "true" ->
+            NewNewDoc = fabric_attachments_handler:inline_att_store(Db, NewDoc); %store_single_document
+		"false" ->
+            couch_log:debug("Store inline attachmets in Swift disabled",[]),
+            NewNewDoc = NewDoc
```
kxepal: Better not assign variable in case, but assign case result to variable.

src/chttpd_db.erl
```diff
@@ -733,7 +741,17 @@ db_doc_req(#httpd{method='POST', user_ctx=Ctx}=Req, Db, DocId) ->
     NewDoc = Doc#doc{
         atts = UpdatedAtts ++ OldAtts2
     },
-    case fabric:update_doc(Db, NewDoc, Options) of
+    couch_log:debug("chttpd_db: couchNew doc formed including new att: formed",[]),
+    lists:foreach(fun(X) -> couch_log:debug("~p~n ~p~n",[couch_att:fetch(data, X), couch_att:fetch(name, X)]) end, NewDoc#doc.atts),
```
kxepal: Please, don't do this: if document has a lot of attachments, this loop will spin without producing any useful work, but only slow down all around.

src/chttpd_db.erl
```diff
@@ -1127,12 +1155,20 @@ db_attachment_req(#httpd{method='GET',mochi_req=MochiReq}=Req, Db, DocId, FileNa
...
         true ->
             fun couch_att:foldl/3
         end,
+        case AttExternal of
+            "external" ->
+                AttFun = couch_att:fetch(data,Att),
+                couch_log:debug("chtttpd_db: Got new attachment ",[]);
+            _ ->
+                couch_log:debug("chttpd_db: not external att",[]),
+                AttFun = AttFunTmp
```
kxepal: So, AttFun could be actually an attachment data or some fold function? Something went wrong if you have to do this.

kxepal:
Is it possible to decouple logic of working with external attachments from the logic that works with internal ones? Mixing them makes quite hard to follow the track.

gilv:
@kxepal : you asked to "Is it possible to decouple logic of working with external attachments from the logic that works with internal ones? Mixing them makes quite hard to follow the track."

Can you please explain me what would you like to see exactly? Perhaps i didn't understand it.
Currently fabric_attachment_handler is the one that responsible to store attachments externally. This module does't really care if it's internal or external, they both go to the same object store.

kxepal:
@gilv I mean to get rid all `case AttExternal of "external" -> ...; _ -> ... end` and instead have some function(s) with two clauses: one to process attachments stored inside db file and other - in some external storage. This makes logic of processing internal and external attachments separated, clean and easy to track. This also helps to avoid situations like [this one](https://github.com/apache/couchdb-chttpd/pull/82/files#r41985231) when depending on attachment type.

gilv:
@kxepal @rnewson 
I provided another patch, implementing most of the comments.
Here is what is not implemented in this patch:

1.  Remark on the AttFun. Not yet implemented. Original remark:  “AttFun could be actually an attachment data or some fold function? Something went wrong if you have to do this.” 
2.  usage of  couch_util:to_hex
3.  still no unitests

rnewson:
Given the nature of the patch, this cannot merge without tests, assuming all style and other issues are resolved.

src/chttpd_db.erl
```diff
@@ -407,7 +407,16 @@ db_req(#httpd{method='POST',path_parts=[_,<<"_bulk_docs">>], user_ctx=Ctx}=Req,
...
-        case fabric:update_docs(Db, Docs, Options2) of
+        case fabric_att_handler:external_store() of
+            true ->
+                couch_log:debug("store attachment externaly", []),
+                NewDocs = fabric_att_handler:att_store(Db,
+                                                          fabric:docs(Docs));
+            false ->
+                couch_log:debug("externalize attachment: disabled",[]),
+                NewDocs = Docs
+        end,
```
rnewson: It would be cleaner if the result of the case statement was assigned to `NewDocs`. i.e, `NewDocs = case...`

rnewson: hm, and really, it should be `fabric:update_docs` itself that makes this decision, the http presentation layer should not know.

src/chttpd_db.erl
```diff
@@ -722,7 +731,14 @@ db_doc_req(#httpd{method='POST', user_ctx=Ctx}=Req, Db, DocId) ->
     NewDoc = Doc#doc{
         atts = UpdatedAtts ++ OldAtts2
     },
-    case fabric:update_doc(Db, NewDoc, Options) of
+    NewNewDoc = case fabric_att_handler:external_store() of
+        true ->
+            fabric_att_handler:att_store(Db, NewDoc); %store_single_document
+		false ->
```
rnewson: indentation

src/chttpd_db.erl
```diff
@@ -722,7 +731,14 @@ db_doc_req(#httpd{method='POST', user_ctx=Ctx}=Req, Db, DocId) ->
...
+            couch_log:debug("Store inline attachmets in Swift disabled",[]),
```
rnewson: this mentions Swift when it shouldn't (it might be stored there)

rnewson:
summary: there's too much logic in the http layer, fabric should be the place where the decisions are made.

gilv:
@kxepal I need your advice on the following:
Consider the following use case:
1. Attachment is stored externally
2. User download attachment
3. CouchDB fails to download attachment from Swift ( for example network failure )

The relevant code in "chttpd_db" is:

ExtStoreID = couch_att:fetch(att_extstore_id,TmpAtt),
        case ExtStoreID of
            undefined ->
                couch_log:debug("chttpd_db: Att is not stored in the external store ~p~n",[ExtStoreID]),
                Att = TmpAtt;
            _ ->
                Att = fabric_att_handler:att_get(Db, TmpAtt),
                couch_log:debug("chtttpd_db: Got attachment from the external store ~p~n",[ExtStoreID])
        end,

My question is: what is the best way to handle this? 
Should i do something like this:

case   fabric_att_handler:att_get(Db, TmpAtt),
          {ok, Att} ->
                couch_log:debug("chtttpd_db: Got attachment from the external store ~p~n",[ExtStoreID])
         {error,_} ->
                throw({
                        bad_request,
                        "Please try again later"
                    })
.....

kxepal:
@gilv It's not a bad request since it's not a user fault that Swift is down. HTTP 503 or HTTP 502 looks good here. But yes, the only thing you can do is to fail and throw error back.

Btw, what's the timeout you use for fetching remote attachments? Do you stream response from Swift or buffer it in memory before return back to user?

gilv:
@kxepal @rnewson 
Few notes related recent patch
1. Please ignore all debug / info print statements. At some point i will remove them at once.
2. I addressed most of the comments.
3. Remark on the AttFun. Not yet implemented. Original remark: “AttFun could be actually an attachment data or some fold function? Something went wrong if you have to do this.”
4. Still no unitests / functional tests.
5. Attachment retrieval from the object store - keep all data in memory. I need to figure out how to use streaming there.

gilv:
@kxepal Can you please review the recent code?

PR #83, opened 2015-10-19
Handle bad replicate request

jaydoane:
This patch is correctly handles https://github.com/apache/couchdb-couch-replicator/pull/12, and obsoletes https://github.com/cloudant/chttpd-local/pull/15

BugsID: 48602


kxepal:
@jaydoane you may close your PR since we did this in a bit better way https://github.com/apache/couchdb-chttpd/commit/33a242edd90f63c5ed115e67c08b7bab640eedc2

jaydoane:
@kxepal, ok closing.

Should I rework https://github.com/apache/couchdb-couch-replicator/pull/12/files so it works with this new code?

kxepal:
@jaydoane I think there is no need. Eventually, it should just work, but check again to be sure. Also, you can rebase it so I'll do fast forward merge - that would be helpful (:

PR #86, opened 2015-10-27
Update /_replicate handler and clean it from legacy decisions

kxepal:
It's not actual anymore to have try/catch around replicate/2 as it never
throws any errors making catch block is actually unreached for
the specified clauses.

We also shouldn't repeat our self in error response rendering here.


rnewson:
+1

PR #89, opened 2015-10-29
Fix Last-Event-ID header handling for db changes feed

kxepal:
This assumes Last-Event-ID header value to be opaque.

COUCHDB-2858


rnewson:
+1

PR #90, opened 2015-11-01
Fix view ETag calculation

KlausTrainer:
COUCHDB-2859

### DIFF
--- a/src/chttpd_view.erl
+++ b/src/chttpd_view.erl
@@ -25,13 +25,13 @@ multi_query_view(Req, Db, DDoc, ViewName, Queries) ->
         couch_mrview_util:validate_args(QueryArg)
     end, Queries),
     {ok, Resp2} = couch_httpd:etag_maybe(Req, fun() ->
-        VAcc0 = #vacc{db=Db, req=Req, prepend="\r\n"},
-        %% TODO: proper calculation of etag
-        Etag = [$", couch_uuids:new(), $"],
+        Etag = make_etag(Db, DDoc, ViewName, ArgQueries),
+        ok = check_etag(Etag, Req),
+        VAcc0 = #vacc{db=Db, req=Req, prepend="\r\n", etag=Etag},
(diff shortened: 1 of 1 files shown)


src/chttpd_view.erl
```diff
@@ -94,3 +94,20 @@ handle_view_req(Req, _Db, _DDoc) ->
 handle_temp_view_req(Req, _Db) ->
     Msg = <<"Temporary views are not supported in CouchDB">>,
     chttpd:send_error(Req, 403, forbidden, Msg).
+
+make_etag(Db, DDoc, ViewName, Args) ->
+    DbName = Db#db.name,
+    DDocId = DDoc#doc.id,
+    {ok, ViewGroupInfo} = fabric:get_view_group_info(DbName, DDocId),
```
rnewson: this is expensive, we must confirm it's worth it. For changing views, I'm skeptical. I'd rather we removed the misleading ETag header.

kxepal: If so...We have some ddoc_cache. Can we have view_group_info_cache for the same purpose?

rnewson: maybe, but remember that caching isn't magic, it's only useful if it's cheaper than querying again.

Here's an implementation idea that should be no more expensive than what we do today;

1) delay the response until we have heard from one shard of each range (I believe we already added this, but do check)
2) from this, construct the etag (we can change the workers to return more information along with their first rows (if any)
3) send 304 if the etag matches the header

That is, we have to hear from one shard per range to form the view, so we can build a useful etag from that.

For 2.0, I would remove the random etag without replacement, since that is probably causing useless cache bloat on the receiving end, we implicitly promise there's a chance of a cache hit where there is no chance.

KlausTrainer: Thanks all for your consideration. I'm closing this in favour of a new pull request that removes the random ETag.

I like @rnewson's implementation idea, and I'm also happy to help implementing it.

PR #91, opened 2015-11-02
Explicitly create admin user and authenticate requests in chttpd_db_test cases

jaydoane:
This patch explicitly configures an admin user and then makes all requests using that user's credentials. This approach continues to work under the CouchDB's security model, and also allows the tests to pass under Cloudant's security model.

COUCHDB-2865
BugzID: 55502

### DIFF
--- a/test/chttpd_db_test.erl
+++ b/test/chttpd_db_test.erl
@@ -15,8 +15,13 @@
 -include_lib("couch/include/couch_eunit.hrl").
 -include_lib("couch/include/couch_db.hrl").
 
+-define(USER, "chttpd_db_test_admin").
+-define(PASS, "pass").
+-define(AUTH, {basic_auth, {?USER, ?PASS}}).
+-define(CONTENT_JSON, {"Content-Type", "application/json"}).
 
 setup() ->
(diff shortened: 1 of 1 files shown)


kxepal:
LGFM, though this isn't the only place where we do HTTP requests from Erlang to CouchDB API and you may need more generalized way to define ?AUTH. But let's postpone this till the next need in this.

Btw, iirc, branch name shouldn't contain bugzid, but jira one (not sure about commit message). Is it all ok with Cloudant policies?

jaydoane:
@kxepal, thanks for the review. I believe you are correct that there are other tests which also rely on implicit "admin party" authorization, and I will tackle those as I come to them.

I also confirmed with Robert Newson that you are correct about the branch name and reference. I was under the mistaken impression that I should open an issue in each bug tracker (since from Cloudant's perspective, this is a bug, but from CouchDB's perspective, it's an enhancement). However, moving forward, I will just use JIRA for issues such as this.

I don't have a commit bit, so can you merge this for me?

kxepal:
@jaydoane Landed. Since you have JIRA issue ticket and traceback magic word, we're fine here (: Thanks!

PR #92, opened 2015-11-02
Don't set random ETag header for views

KlausTrainer:
This fixes the issue that by setting a random ETag header, we implicitly
promised that there's a chance of a cache hit where there was no chance.

For 2.0, we simply remove the random ETag without replacement, in order
to not accidentally cause useless cache bloat in clients.

Here's Robert Newson's idea for an implementation that should be no more
expensive than what we do today:

1) delay the response until we have heard from one shard of each range
2) from this, construct the etag (we can change the workers to return
   more information along with their first rows (if any))
3) send 304 if the etag matches the header

That is, we have to hear from one shard per range to form the view, so
we can build a useful etag from that.

COUCHDB-2859

### DIFF
--- a/src/chttpd_view.erl
+++ b/src/chttpd_view.erl
@@ -24,41 +24,23 @@ multi_query_view(Req, Db, DDoc, ViewName, Queries) ->
         QueryArg = couch_mrview_http:parse_params(Query, undefined, Args1),
         couch_mrview_util:validate_args(QueryArg)
     end, Queries),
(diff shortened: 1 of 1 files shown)


kxepal:
+1. But lets move most part of the commit message to JIRA issue for the better Etag impl for views. Basically, here we remove random Etag set because it's useless, wrong and misleading.

KlausTrainer:
Thanks, @kxepal. I amended the commit message and added Robert's idea to [COUCHDB-2859](https://issues.apache.org/jira/browse/COUCHDB-2859).

PR #93, opened 2015-11-02
Don't set random ETag header for `_all_docs` view

KlausTrainer:
This fixes the issue that by setting a random ETag header, we implicitly
promised that there's a chance of a cache hit where there was no chance.

For 2.0, we simply remove the random ETag without replacement, in order
to not accidentally cause useless cache bloat in clients.

COUCHDB-2859


kxepal:
No more lottery! +1

KlausTrainer:
Thanks :)

PR #94, opened 2015-11-02
Fix return types of responses to view requests

KlausTrainer:
This fixes an issue that was introduced with 72258e2 and e5b1aee.

Background:

Although all things still seemed to work, each request to a view
generated an exception with the following stack trace:

```
exception error: undefined function vacc:get/2
  in function  chttpd:result/2 (src/chttpd.erl, line 297)
  in call from chttpd:handle_request_int/1 (src/chttpd.erl, line 276)
  in call from mochiweb_http:headers/6 (src/mochiweb_http.erl, line 122)
```

### DIFF
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -618,7 +618,8 @@ all_docs_view(Req, Db, Keys, OP) ->
     Options = [{user_ctx, Req#httpd.user_ctx}],
     Max = chttpd:chunked_response_buffer_size(),
     VAcc = #vacc{db=Db, req=Req, threshold=Max},
-    fabric:all_docs(Db, Options, fun couch_mrview_http:view_cb/2, VAcc, Args3).
+    {ok, Resp} = fabric:all_docs(Db, Options, fun couch_mrview_http:view_cb/2, VAcc, Args3),
+    {ok, Resp#vacc.resp}.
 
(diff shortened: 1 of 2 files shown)


kxepal:
/:

KlausTrainer:
Well, I'm also :/ on this. I'd love to understand why the corresponding requests would succeed anyway, but don't have time to dig deep enough for that right now.

kxepal:
May be:
```
    case is_record(Resp2, vacc) of
        true -> {ok, Resp2#vacc.resp};
        _ -> {ok, Resp2}
    end.
```
Had some meaning there and it worth to restore it?

KlausTrainer:
At least I'm quite sure about this one…

Previously this was needed because `couch_httpd:etag_maybe/2` was called there. Its return value contains either a mochiweb response (if the etag matches), or a `vacc` record (if the etag doesn't match).

Now that we don't call `couch_httpd:etag_maybe/2` there anymore, the response should always be a `vacc` record, and therefore the case clause should be obsolete.

kxepal:
Tested now. Looks as everything works, so +1 for the fix, but you brought [COUCHDB-2650](https://issues.apache.org/jira/browse/COUCHDB-2650). What shall we do? Seems like there is no way, but to implement Etag property for 2.0 (:

PR #95, opened 2015-11-18
Handle errors from before_request/after_request

iilyak:
This refactoring fixes the handling of errors from before_request.
Unify how we deal with errors among before_request and handle_request.

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -290,12 +290,11 @@ catch_error(HttpReq, exit, {mochiweb_recv_error, E}) ->
         peer = Peer,
         original_method = Method
     } = HttpReq,
-    LogForClosedSocket = io_lib:format("mochiweb_recv_error for ~s - ~p ~s", [
+    couch_log:notice("mochiweb_recv_error for ~s - ~p ~s - ~p", [
         Peer,
         Method,
-        MochiReq:get(raw_path)
-    ]),
-    couch_log:notice(LogForClosedSocket ++ " - ~p", [E]),
+        MochiReq:get(raw_path),
+        E]),
     exit(normal);
 catch_error(HttpReq, exit, {uri_too_long, _}) ->
     send_error(HttpReq, request_uri_too_long);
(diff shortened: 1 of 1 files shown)


src/chttpd.erl
```diff
@@ -242,49 +274,47 @@ handle_request_int(MochiReq) ->
...
+catch_error(HttpReq, error, database_does_not_exist) ->
+    send_error(HttpReq, database_does_not_exist);
+catch_error(HttpReq, Tag, Error) ->
+    Stack = erlang:get_stacktrace(),
+    % TODO improve logging and metrics collection for client disconnects
+    case {Tag, Error, Stack} of
+        {exit, normal, [{mochiweb_request, send, _, _} | _]} ->
+            exit(normal); % Client disconnect (R15+)
+        {exit, normal, [{mochiweb_request, send, _} | _]} ->
+            exit(normal); % Client disconnect (R14)
```
kxepal: You can drop this clause with the light on heart like we dropped R14 support.

src/chttpd.erl
```diff
@@ -242,49 +274,47 @@ handle_request_int(MochiReq) ->
...
+        mochi_req = MochiReq,
+        peer = Peer,
+        original_method = Method
+    } = HttpReq,
+    LogForClosedSocket = io_lib:format("mochiweb_recv_error for ~s - ~p ~s", [
+        Peer,
+        Method,
+        MochiReq:get(raw_path)
+    ]),
+    couch_log:notice(LogForClosedSocket ++ " - ~p", [E]),
```
kxepal: Why not to do formatting right in couch_log:notice call?

src/chttpd.erl
```diff
@@ -209,21 +204,58 @@ handle_request_int(MochiReq) ->
...
+    erlang:put(dont_log_request, true),
+
+    Result0 = case before_request(HttpReq0) of
+        {ok, HttpReq1} ->
+            process_request(HttpReq1);
+        {error, Response} ->
+            {HttpReq0, Response}
+    end,
+
+    {HttpReq3, HttpResp0} = result(Result0, HttpReq0),
```
kxepal: The way how we form Result0 and how result/2 works very confusing. Especially for the case when before_request returns {error, Response}, because Response have to be {ok, #httpd{}}  or {aborted, MochiReq, Reason} or you'll get function_clause.

Is it possible to make it more friendly for this refactor?

kxepal:
Looks good for me, need to check however.

kxepal:
Nice! +1

PR #100, opened 2016-01-25
Improve cors support

iilyak:
This PR makes it possible to configure CORS to be able to use custom headers.
It mostly fixes already existing code. The only new addition is `exposed_headers` setting.

The supported configuration is via `default.ini`:
```ini
[cors]
headers = "content-type, accept-ranges, ..."
methods = "COPY, DELETE, ...."
exposed_headers = "extra, accept-ranges, etag, server, ..."
```

### DIFF
--- a/src/chttpd_cors.erl
+++ b/src/chttpd_cors.erl
@@ -279,7 +279,7 @@ get_cors_config(#httpd{cors_config = undefined}) ->
         undefined ->
             ?SUPPORTED_HEADERS;
         AllowHeaders0 ->
-            split_list(AllowHeaders0)
+            [to_lower(H) || H <- split_list(AllowHeaders0)]
     end,
     AllowMethods = case config:get("cors", "methods", undefined) of
         undefined ->
@@ -291,7 +291,7 @@ get_cors_config(#httpd{cors_config = undefined}) ->
         undefined ->
             ?COUCH_HEADERS;
         ExposedHeaders0 ->
-            split_list(ExposedHeaders0)
+            [to_lower(H) || H <- split_list(ExposedHeaders0)]
     end,
     Origins0 = binary_split_list(config:get("cors", "origins", [])),
     Origins = [{O, {[]}} || O <- Origins0],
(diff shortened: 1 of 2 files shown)


eiri:
All right, that took some effort to test, I've ended up writing a simple CORS app, but now I'm sure this is working as advertised. 

+1

eiri:
A note: config parameters for `headers` and `exposed_headers` **have to** be in lower case and it's notoriously hard to figure out, browser will just keep telling that OPTION throws 401 error code.

We'd better to reflect this in documentation to protect users.

src/chttpd_cors.erl
```diff
@@ -176,15 +180,15 @@ headers(Req, RequestHeaders, Origin, Config) ->
         true ->
             AcceptedOrigins = get_accepted_origins(Req, Config),
             CorsHeaders = handle_headers(Config, Origin, AcceptedOrigins),
-            maybe_apply_headers(CorsHeaders, RequestHeaders);
+            maybe_apply_headers(Config, CorsHeaders, RequestHeaders);
```
eiri: Personally I'd get `exposed_headers` out of config on a line above to make input parameters in `maybe_apply_headers` consistent to be all headers, but this is a question of a personal taste, so feel free to ignore.

kxepal:
@eiri I think that's not good to ask write methods in upper case and headers in lower one. Users who read the spec knows that header names are case insensitive, so it's not what they expect from us here. Reference to the docs wouldn't work here....well, it will, but after someone waste his time on debug.

Can we take the responsibility for headers names normalization from user to our code?

eiri:
@kxepal Well, you right, I guess we can do `[string:to_lower(H) || H <- split_list(Headers0)]` for `AllowMethods` and `ExposedHeaders`, that'd save someone from some grey hair.

kxepal:
@eiri sounds good!

kxepal:
@iilyak Thank you!
@eiri Does the latest changes works for your test app? Without grey hair now I guess? (:

eiri:
Too late, the harm has been done. :older_man:

This looks good, again: +1 after squash

iilyak:
Squashed.

kxepal:
Once Eric was just Eric, but now he'll be known as Eric the Grey! P:
Good to know, will trust you here as I'm not good with CORS thing. Looks good for me as well. +1

PR #102, opened 2016-02-18
Support array sequence format

jaydoane:
Allow test to pass with JSON sequence format like:
"seq": [num, binary]

### DIFF
--- a/test/chttpd_db_test.erl
+++ b/test/chttpd_db_test.erl
@@ -83,17 +83,19 @@ should_accept_live_as_an_alias_for_continuous(Url) ->
         {ok, _, _, ResultBody} =
             test_request:get(Url ++ "/_changes?feed=live&timeout=1", [?AUTH]),
         {ResultJson} = ?JSON_DECODE(ResultBody),
-        <<LastSeqNum0:1/binary, "-", _/binary>> = couch_util:get_value(
-            <<"last_seq">>, ResultJson, undefined),
-        LastSeqNum = list_to_integer(binary_to_list(LastSeqNum0)),
+        LastSeqNum = seq_num(couch_util:get_value(<<"last_seq">>, ResultJson)),
 
         {ok, _, _, _} = create_doc(Url, "testdoc2"),
         {ok, _, _, ResultBody2} = 
             test_request:get(Url ++ "/_changes?feed=live&timeout=1", [?AUTH]),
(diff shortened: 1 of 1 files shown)


test/chttpd_db_test.erl
```diff
@@ -83,17 +83,19 @@ should_accept_live_as_an_alias_for_continuous(Url) ->
...
         {[{_, Seq}, _]} = ?JSON_DECODE(CleanedResult),
-        <<SeqNum0:1/binary, "-", _/binary>> = Seq,
-        SeqNum = list_to_integer(binary_to_list(SeqNum0)),
+        SeqNum = seq_num(Seq),
 
         ?assertEqual(LastSeqNum + 1, SeqNum)
     end).
+
+seq_num(<<SeqNumBin:1/binary, "-", _/binary>>) ->
+    list_to_integer(binary_to_list(SeqNumBin));
```
kxepal: it's possible now to use `binary_to_integer` as our minimal supported Erlang version is R16B03-1 which has it.

jaydoane:
@kxepal, thanks for the tip. If you are inclined to accept this PR, I would be happy to open a Jira ticket for it.

kxepal:
I think I do (after squash), no need for JIRA for such trivial changes. 
The only question is when array sequence format happens? And why we're not stable on seq format?

jaydoane:
The only place I know that an array is used for seq format is in a soon-to-be legacy version of Cloudant Local. @mikewallace1979 probably has the backstory on where it originally came from.

I can understand any reluctance to adding something like this to couchdb, but the alternatives for keeping downstream testing green appear rather distasteful. My plan was to open this PR to get Mike's opinion on whether we should even pursue the idea of getting this merged upstream, but kxepal beat me to it.

mikewallace1979:
@kxepal @jaydoane Yes we standardized on the string format in https://github.com/apache/couchdb-fabric/commit/42a67d0a018868d3760df90eca9d2e8e10614fd7 and the array format is only required for a downstream version which is nothing to do with the ASF.

That being the case I don't think we should merge this here.

kxepal:
@mikewallace1979 I see. Agree, there is no point in this support then. I think there is nothing wrong for downstream to apply this patch and handle further changes via simple merge till this wouldn't be required.

Otherwise we'll end with a lot odd-and-legacy branches that nobody knows for what they have to be used.

jaydoane:
Makes sense to me! Closing.

PR #103, opened 2016-02-25
Use new optional fetch logic for configure_filter/5

chewbranca:


chewbranca:
Companion PR out in https://github.com/apache/couchdb-couch/pull/144.

chewbranca:
Closing this PR as an alternative implementation was found that makes this change unnecessary.

PR #108, opened 2016-03-17
Use couch_tests applications for chttpd_plugin_tests

iilyak:
This fixes the problem with chttpd_plugin_tests. Previously following would fail:
```
make eunit apps=chttpd tests=callback_test,error_info_test
```
In order to test this PR you would need to add https://github.com/apache/couchdb-erlang-tests/pull/1 as a dependency in `rebar.config.script`. You might need to add the app into `rel/reltool.config` as well.

### DIFF
--- a/test/chttpd_plugin_tests.erl
+++ b/test/chttpd_plugin_tests.erl
@@ -43,16 +43,14 @@ data_subscriptions() -> [].
 processes() -> [].
 notify(_, _, _) -> ok.
 
-setup() ->
-    application:stop(couch_epi), % in case it's already running from other tests...
-    application:unload(couch_epi),
-    ok = application:load(couch_epi),
-    ok = application:set_env(couch_epi, plugins, [?MODULE]), % only this plugin
-    ok = application:start(couch_epi).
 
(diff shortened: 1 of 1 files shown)


test/chttpd_plugin_tests.erl
```diff
@@ -43,16 +43,14 @@ data_subscriptions() -> [].
...
-    ok = application:load(couch_epi),
-    ok = application:set_env(couch_epi, plugins, [?MODULE]), % only this plugin
-    ok = application:start(couch_epi).
-
-teardown(_) ->
-    ok = application:stop(couch_epi),
-    ok = application:unload(couch_epi).
+    couch_tests:setup([
+        couch_epi_dispatch:dispatch(chttpd, ?MODULE)
+    ], [], []).
```
kxepal: Feels like this has a potential for further simplification:
```
couch_tests:setup([couch_epi_dispatch:dispatch(chttpd, ?MODULE)])
```

or even:

```
couch_tests:setup_epi(?MODULE, [chttpd])
```

kxepal:
+1

PR #109, opened 2016-03-21
2973 log user

iilyak:
Change the log format from `chttpd` to include username.

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -1145,4 +1145,46 @@ check_url_encoding_fail_test_() ->
             check_url_encoding("/dbname%g2"))
     ].
 
+log_format_test() ->
+    ?assertEqual(
+        "nonce 127.0.0.1 127.0.0.1:15984 undefined "
+        "GET /_cluster_setup 201 ok 10000",
+        test_log_request("/_cluster_setup", undefined)),
+    ?assertEqual(
+        "nonce 127.0.0.1 127.0.0.1:15984 user_foo "
+        "GET /_all_dbs 201 ok 10000",
+        test_log_request("/_all_dbs", #user_ctx{name = <<"user_foo">>})),
+
+    %% Utf8Name = unicode:characters_to_binary(Something),
+    Utf8User = <<227,130,136,227,129,134,227,129,147,227,129,157>>,
+    ?assertEqual(
+        "nonce 127.0.0.1 127.0.0.1:15984 %E3%82%88%E3%81%86%E3%81%93%E3%81%9D "
+        "GET /_all_dbs 201 ok 10000",
+        test_log_request("/_all_dbs", #user_ctx{name = Utf8User})),
+    ok.
+
+test_log_request(RawPath, UserCtx) ->
+    Headers = mochiweb_headers:make([{"HOST", "127.0.0.1:15984"}]),
+    MochiReq = mochiweb_request:new(socket, [], 'POST', RawPath, version, Headers),
+    Req = #httpd{
+        mochi_req = MochiReq,
+        begin_ts = {1458,588713,124003},
+        original_method = 'GET',
+        peer = "127.0.0.1",
+        nonce = "nonce",
+        user_ctx = UserCtx
+    },
+    Resp = #httpd_resp{
+        end_ts = {1458,588723,124303},
+        code = 201,
+        status = ok
+    },
+    ok = meck:new(couch_log, [passthrough]),
+    ok = meck:expect(couch_log, notice, fun(Format, Args) ->
+        lists:flatten(io_lib:format(Format, Args))
+    end),
+    Message = maybe_log(Req, Resp),
+    ok = meck:unload(couch_log),
+    Message.
+
 -endif.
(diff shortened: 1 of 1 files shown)


src/chttpd.erl
```diff
@@ -1074,3 +1075,53 @@ respond_(#httpd{mochi_req = MochiReq}, Code, Headers, _Args, start_response) ->
     MochiReq:start_response({Code, Headers});
 respond_(#httpd{mochi_req = MochiReq}, Code, Headers, Args, Type) ->
     MochiReq:Type({Code, Headers, Args}).
+
+get_user(#httpd{user_ctx = #user_ctx{name = User}}) -> User;
+get_user(#httpd{user_ctx = undefined}) -> "undefined".
```
kxepal: What's the case for undefined? I thought we always have user_ctx set...

kxepal: Nvm, I see it. Thought it's strange that this value is unstable depending on what operation you proceed.

kxepal:
+1

kxepal:
May be one addition to not log user names as UTF-8, but in some JSON to not get fooled by unicode tricky usernames or get log lines broken by newlines in username.

kxepal:
Oh yes, user names `"null"` and  `"undefined"` will cause a lot of fun.

iilyak:
@kxepal Supporting utf-8 is a very valid concern. Do we actually support it? It doesn't seem possible to transmit utf-8 in a http header. We use basic auth which is based on headers. There is a new [RFC7617](https://datatracker.ietf.org/doc/rfc7617/) which is going to support utf-8. But currently it is not supported. Therefore I do believe that we shouldn't have any utf-8 users in the wild. As utf-8 support is slowly coming maybe we should consider userid validation on couch side to sanitize user's input before it became a problem. The proposed [RFC7613](https://datatracker.ietf.org/doc/rfc7613/?include_text=1) defines what can be in a userid and what shouldn't be there.

iilyak:
@kxepal: I did create a jira ticket [COUCHDB-2974](https://issues.apache.org/jira/browse/COUCHDB-2974) to track utf-8 support.

iilyak:
> @kxepal: Oh yes, user names "null" and "undefined" will cause a lot of fun.

Would you recommend changing `"undefined"` -> `"null"`?

kxepal:
>>Oh yes, user names "null" and "undefined" will cause a lot of fun.

> Would you recommend changing "undefined" -> "null"?

No, I mean that user with name "null" will looks in logs as the anonymous user - so you'll never trace him. Same story with "undefined". I don't think there is any solution except to quote the name or do something to distinguish authed user from anonymous ones.

kxepal:
As for UTF-8: yes, we do. More over, we already support UTF-8 in basic auth header regardless the fact that RFC assumes ASCII there.

iilyak:
@kxepal: Should we log `hex(base64(name))` instead then? Even ASCII could have `\a` or `\r` which could brake the log lines. It such a pity that we historically do not validate user names.

kxepal:
@iilyak I would prefer json or url encoded way. Both are fine to notice tricky bits, but each is useful for further copy-paste into some script to work with it and both provides more-or-less readable close to real variants of username. Plain ASCII ones will remain plain ASCII ones.

However, that could be not very friendly for the end users (admins/devops) to work with such kind of data. Need to think about and call for more opinions around. For instance, what tools ElasticSearch could provide to work with such kind of encoded usernames?

rnewson:
we could apply url encoding before logging it. That way administrators can decode to the real value but we don't have to worry about any unicode trickery.

kxepal:
+1 url encoding then. That's better than json as you can simply copy-paste the name to url and get the doc, it's still readable for ASCII names and quite easy to convert into UTF-8 string if need.

src/chttpd.erl
```diff
@@ -1074,3 +1075,60 @@ respond_(#httpd{mochi_req = MochiReq}, Code, Headers, _Args, start_response) ->
     MochiReq:start_response({Code, Headers});
 respond_(#httpd{mochi_req = MochiReq}, Code, Headers, Args, Type) ->
     MochiReq:Type({Code, Headers, Args}).
+
+get_user(#httpd{user_ctx = #user_ctx{name = User}}) ->
+    mochiweb_util:quote_plus(User);
```
rnewson: couch_util:url_encode ?

rnewson:
+1

src/chttpd.erl
```diff
@@ -1074,3 +1075,60 @@ respond_(#httpd{mochi_req = MochiReq}, Code, Headers, _Args, start_response) ->
...
+    ok = meck:new(couch_log, [passthrough]),
+    ok = meck:expect(couch_log, notice, fun(Format, Args) ->
+        lists:flatten(io_lib:format(Format, Args))
+    end),
+    Message = maybe_log(Req, Resp),
+    ok = meck:unload(couch_log),
+    Message.
+
+
+
```
kxepal: Reserved for future tests? (:

kxepal:
+1, but need to rebase. Bonus +1 if empty space will get reduced (;

PR #111, opened 2016-04-07
Create md5 etag for _local docs

garrensmith:
This makes a unique ETAG for _local docs, so that they are cached
correctly, and fetched again when the document changes.

fixes COUCHDB-2978

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -621,8 +621,8 @@ json_body_obj(Httpd) ->
     end.
 
 
-doc_etag(#doc{revs={Start, [DiskRev|_]}}) ->
-    "\"" ++ ?b2l(couch_doc:rev_to_str({Start, DiskRev})) ++ "\"".
+doc_etag(#doc{id=Id, body=Body, revs={Start, [DiskRev|_]}}) ->
+    couch_httpd:doc_etag(Id, Body, {Start, DiskRev}).
 
 make_etag(Term) ->
     <<SigInt:128/integer>> = couch_crypto:hash(md5, term_to_binary(Term)),
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -950,7 +950,7 @@ http_code_from_status(Status) ->
             200
     end.
 
-update_doc(Db, DocId, #doc{deleted=Deleted}=Doc, Options) ->
+update_doc(Db, DocId, #doc{deleted=Deleted, body=DocBody}=Doc, Options) ->
     {_, Ref} = spawn_monitor(fun() ->
         try fabric:update_doc(Db, Doc, Options) of
             Resp ->
@@ -981,8 +981,7 @@ update_doc(Db, DocId, #doc{deleted=Deleted}=Doc, Options) ->
     {accepted, NewRev} ->
         Accepted = true
     end,
-    NewRevStr = couch_doc:rev_to_str(NewRev),
-    Etag = <<"\"", NewRevStr/binary, "\"">>,
+    Etag = couch_httpd:doc_etag(DocId, DocBody, NewRev),
     Status = case {Accepted, Deleted} of
         {true, _} ->
             accepted;
@@ -991,6 +990,7 @@ update_doc(Db, DocId, #doc{deleted=Deleted}=Doc, Options) ->
         {false, false} ->
             created
     end,
+    NewRevStr = couch_doc:rev_to_str(NewRev),
     Body = {[{ok, true}, {id, DocId}, {rev, NewRevStr}]},
     {Status, {etag, Etag}, Body}.
 


kxepal:
I was thought that the idea is to not send Etag for _local docs at all. Now it breaks the rule that Etag for docs matches _rev field value.

garrensmith:
@rnewson and I figured sending an md5 hash of the body and the revision might be better than not sending anything.

rnewson:
Where do we promise that ETag is the _rev value? We have X-CouchDB-NewRev when we intend to convey the new rev value, for example.

I like this change, it makes _local doc ETags work correctly. Clients can cache based on this and their cache hits will be for the correct body content (that's why rev and body are in the calculation, as both appear in the JSON response body but are stored in separate fields of #doc{}).

kxepal:
@rnewson Everywhere all the time:

```
$ echo '{"foo":4}' | http put http://localhost:15984/db/test
HTTP/1.1 201 Created
Content-Length: 67
Date: Thu, 07 Apr 2016 14:13:26 GMT
Etag: "1-80d7ea9795ad67ad0e7a1e567c15314e"
Location: http://localhost:15984/db/test
Server: CouchDB/fc9acc2 (Erlang OTP/17)
X-Couch-Request-ID: 13cab4abd5
X-CouchDB-Body-Time: 0

{
    "id": "test",
    "ok": true,
    "rev": "1-80d7ea9795ad67ad0e7a1e567c15314e"
}
```

```
$ echo '{"foo":4, "_rev": "1-ABC"}' | http put http://localhost:15984/db/test?new_edits=false
HTTP/1.1 201 Created
Content-Length: 38
Date: Thu, 07 Apr 2016 14:13:39 GMT
Etag: "1-ABC"
Location: http://localhost:15984/db/test
Server: CouchDB/fc9acc2 (Erlang OTP/17)
X-Couch-Request-ID: 2c37ce6ed3
X-CouchDB-Body-Time: 0

{
    "id": "test",
    "ok": true,
    "rev": "1-ABC"
}
```

Also [from docs](http://docs.couchdb.org/en/latest/api/document/common.html#head--db-docid). And many entries in wiki I'm lazy to search for. That's an old feature that simplifies caching without adding some new entity to remember.

kxepal:
As for ` X-CouchDB-NewRev `, this only used for update functions and as a workaround the fact, that it's impossible to return a new revision for the updated doc from there.

rnewson:
I'm ok with changing this for _local docs.

rnewson:
+1

rnewson:
@kxepal thanks for the doc link. I suggest we update to say it doesn't apply to _local docs, which nothing would usually cache, because it's only the replicator (or libraries implementing replication) that would get/put them.

kxepal:
@rnewson @garrensmith 
May I ask you about the reasons to include Etag if _local docs are not intended to be cached? Not for dispute, but to clarify that moment and know what to answer on "wtf why so inconsistent"? (:

P.S. However, with PouchDB we could say, that every web browser now is a replicator, so eventually, they'll use that feature (and that's why this whole story get started).

kxepal:
I guess, I figured why and how this could be explained: we don't use _rev for _local docs as they are not tracked by MVCC feature, but we still want to provide some sane token for caching and notifying the client that the content had changed. So Etag forms using the same rules as the _rev, but doc itself doesn't uses MVCC _rev feature. Sounds good for me.

+1

rnewson:
right. the ETag is to allow correct caching, and I see no reason to make them uncacheable when we can easily mix in the body. Also, there are browsers that do caching poorly in the absence of ETag, so this should fix a bug or two.

cc @daleharvey @nolanlawson

nolanlawson:
I'll note that WebKit has actually fixed the bug in master, so it should appear in the next Safari release: https://bugs.webkit.org/show_bug.cgi?id=156048

In general, though, I'm still +1 on this PR because it's best to not make assumptions about subtleties of browser caching. Firefox or Chrome could change their implementation tomorrow and break in a new way. This ETag behavior LGTM.

daleharvey:
Yeh the behaviour of Etag doesnt match the couchdb semantics for `_rev` and using for the same thing causes bugs, I would really like to see this for all docs (ie if you ever match an Etag, you are guaranteed the same content) but at least having it for local docs would be nice

rnewson:
@daleharvey but the _rev for regular documents mixes in the body and the attachment md5's, this is the first I'm hearing that the ETag for regular docs is wrong.

daleharvey:
In CouchDB it does however you can write documents with any revision you want - https://issues.apache.org/jira/browse/COUCHDB-2764, its a bit of an edge case so I havent pushed it, but still possible

rnewson:
ah true. yes, fair. We'd like to _guarantee_ that the ETag is a digest of the response body (either directly or indirectly).

PR #112, opened 2016-04-12
tolerate failure when inserting auth ddoc

rnewson:
COUCHDB-2986


davisp:
+1 though we'll want to pay attention to reports of errors logged there. We need this to prevent the bug from spidering but this seems like we're not correctly handling an error in the fabric rpc coordinator.

rnewson:
there's a cause of this that's unfound. a fabric:open_doc should not report {error, read_failure} with 1 node down out of 3, and certainly not _continue_ to do so forever.

PR #113, opened 2016-04-12
Reject incorrectly encoded urls

eiri:
This makes chttpd to raise `400 Bad Request` error in case provided URL consist "%" character that not followed by two hex characters as required by RFC 2141.

Related issue: COUCHDB-2748

### DIFF
--- a/include/chttpd.hrl
+++ b/include/chttpd.hrl
@@ -20,3 +20,9 @@
     should_log = true,
     reason
 }).
+
+-define(is_hex(C), (
+    (C >= $0 andalso C =< $9) orelse
+    (C >= $a andalso C =< $f) orelse
+    (C >= $A andalso C =< $F)
+)).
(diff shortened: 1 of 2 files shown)


kxepal:
Nice! LGFM, +1

src/chttpd.erl
```diff
@@ -406,6 +407,15 @@ check_request_uri_length(Uri, MaxUriLen) when is_list(MaxUriLen) ->
             ok
     end.
 
+check_uri_encoding([]) ->
+    ok;
+check_uri_encoding([$%, A, B | Rest]) when ?is_hex(A), ?is_hex(B) ->
+    check_uri_encoding(Rest);
+check_uri_encoding([$% | _]) ->
+    throw({bad_request, invalid_urlencode});
```
rnewson: `invalid_url_encoding`

rnewson: you also switch from `uri` in the function name to `url` in the error value.

include/chttpd.hrl
```diff
@@ -20,3 +20,9 @@
     should_log = true,
     reason
 }).
+
+-define(is_hex(C),
+    (C >= $0 andalso C =< $9) orelse
```
rnewson: do we want a surrounding `()` around the three clauses here so this is safe to include anywhere without unexpected side-effects?

rnewson:
+1 after comment pondering.

PR #114, opened 2016-04-17
Add support for single_max_doc

tonysun83:
Users have been complaining about performance issues as well as indexing timeout issues.  Some of these issues are due to large documents. Currently, our only limitation is the max_document_size parameter that is set to 64 MB. With this enhancement, the goal is to provider some finer grained control over single document sizes and provide some more error info. The max_document_size still takes precedence. The default max size for a single doc is 1 MB.

During bulk requests, when one of the documents exceeds the single_doc_max_size value, it will not be created/updated. The error message will simply be "too_large" for that document. Other documents under the limit will be created/updated normally.

COUCHDB-2992

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -599,12 +599,9 @@ body_length(#httpd{mochi_req=MochiReq}) ->
 body(#httpd{mochi_req=MochiReq, req_body=ReqBody}) ->
     case ReqBody of
         undefined ->
-            % Maximum size of document PUT request body (4GB)
-            MaxSize = list_to_integer(
-                config:get("couchdb", "max_document_size", "4294967296")),
             Begin = os:timestamp(),
             try
-                MochiReq:recv_body(MaxSize)
+                MochiReq:recv_body(?DEFAULT_RECV_BODY)
             after
                 T = timer:now_diff(os:timestamp(), Begin) div 1000,
                 put(body_time, T)
@@ -858,9 +855,9 @@ error_info({error, {database_name_too_long, DbName}}) ->
         <<"At least one path segment of `", DbName/binary, "` is too long.">>};
 error_info({missing_stub, Reason}) ->
     {412, <<"missing_stub">>, Reason};
-error_info({single_doc_too_large, MaxSize}) ->
+error_info({request_entity_too_large, MaxSize}) ->
     SizeBin = list_to_binary(integer_to_list(MaxSize)),
-    {413, <<"too_large">>, <<"Document exceeded single_max_doc_size:",
+    {413, <<"too_large">>, <<"Document exceeded max_document_size:",
         SizeBin/binary>>};
 error_info(request_entity_too_large) ->
     {413, <<"too_large">>, <<"the request entity is too large">>};
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -858,6 +858,10 @@ error_info({error, {database_name_too_long, DbName}}) ->
         <<"At least one path segment of `", DbName/binary, "` is too long.">>};
 error_info({missing_stub, Reason}) ->
     {412, <<"missing_stub">>, Reason};
+error_info({single_doc_too_large, MaxSize}) ->
+    SizeBin = list_to_binary(integer_to_list(MaxSize)),
+    {413, <<"too_large">>, <<"Document exceeded single_max_doc_size:",
+        SizeBin/binary>>};
 error_info(request_entity_too_large) ->
     {413, <<"too_large">>, <<"the request entity is too large">>};
 error_info({error, security_migration_updates_disabled}) ->
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -19,7 +19,7 @@
     db_req/2, couch_doc_open/4,handle_changes_req/2,
     update_doc_result_to_json/1, update_doc_result_to_json/2,
     handle_design_info_req/3, handle_view_cleanup_req/2,
-    update_doc/4, http_code_from_status/1]).
+    update_doc/4, http_code_from_status/1, maybe_verify_body_size/1]).
 
 -import(chttpd,
     [send_json/2,send_json/3,send_json/4,send_method_not_allowed/2,
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -309,10 +309,8 @@ db_req(#httpd{method='POST', path_parts=[DbName], user_ctx=Ctx}=Req, Db) ->
     W = chttpd:qs_value(Req, "w", integer_to_list(mem3:quorum(Db))),
     Options = [{user_ctx,Ctx}, {w,W}],
 
-    Body = chttpd:json_body(Req),
-    ok = maybe_verify_body_size(Body),
-
-    Doc = couch_doc:from_json_obj(Body),
+    Doc = couch_doc:from_json_obj(chttpd:json_body(Req)),
+    ok = maybe_verify_body_size(Doc#doc.body),
     Doc2 = case Doc#doc.id of
         <<"">> ->
             Doc#doc{id=couch_uuids:new(), revs={0, []}};
@@ -769,10 +767,8 @@ db_doc_req(#httpd{method='PUT', user_ctx=Ctx}=Req, Db, DocId) ->
         case chttpd:qs_value(Req, "batch") of
         "ok" ->
             % batch
-            Body = chttpd:json_body(Req),
-            ok = maybe_verify_body_size(Body),
-            Doc = couch_doc_from_req(Req, DocId, Body),
-
+            Doc = couch_doc_from_req(Req, DocId, chttpd:json_body(Req)),
+            ok = maybe_verify_body_size(Doc#doc.body),
             spawn(fun() ->
                     case catch(fabric:update_doc(Db, Doc, Options)) of
                     {ok, _} -> ok;
@@ -788,8 +784,8 @@ db_doc_req(#httpd{method='PUT', user_ctx=Ctx}=Req, Db, DocId) ->
         _Normal ->
             % normal
             Body = chttpd:json_body(Req),
-            ok = maybe_verify_body_size(Body),
             Doc = couch_doc_from_req(Req, DocId, Body),
+            ok = maybe_verify_body_size(Doc#doc.body),
             send_updated_doc(Req, Db, DocId, Doc, RespHeaders, UpdateType)
         end
     end;
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -13,6 +13,7 @@
 -module(chttpd_db).
 -include_lib("couch/include/couch_db.hrl").
 -include_lib("couch_mrview/include/couch_mrview.hrl").
+-include_lib("chttpd/include/chttpd.hrl").
 
 -export([handle_request/1, handle_compact_req/2, handle_design_req/2,
     db_req/2, couch_doc_open/4,handle_changes_req/2,
@@ -309,7 +310,7 @@ db_req(#httpd{method='POST', path_parts=[DbName], user_ctx=Ctx}=Req, Db) ->
     Options = [{user_ctx,Ctx}, {w,W}],
 
     Body = chttpd:json_body(Req),
-    ok = verify_doc_size(Body),
+    ok = maybe_verify_body_size(Body),
 
     Doc = couch_doc:from_json_obj(Body),
     Doc2 = case Doc#doc.id of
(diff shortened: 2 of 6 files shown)


src/chttpd_db.erl
```diff
@@ -1628,6 +1634,34 @@ bulk_get_open_doc_revs1(Db, Props, _, {DocId, Revs, Options}) ->
...
+        end,
+        Reason = lists:concat(["Document exceeded single_max_doc_size:",
+            " of ", MaxSize, " bytes"]),
+        {[{id, DocId}, {error, <<"too_large">>},
+            {reason, ?l2b(Reason)}]}
+    end, ExceedDocs),
+    {ExceedErrs, DocsArray}.
+
+exceed_doc_size(JsonBody, MaxSize) ->
+    size(term_to_binary(JsonBody)) > MaxSize.
```
tonysun83: Is this the accurate method of figuring out the bytes of the Json Body?

kxepal:
Hi Tony! Nice feature, but...

1. Why there is a need to introduce another configuration option instead of reusing existed `max_document_size `? The case: I see in configuration options:

```
[couchdb]
max_document_size = 1048576
single_max_doc_size = 2097152
```

What the document max size will be actually allowed? I guess, without code/experiment that would be hard to say for sure.

P.S. Yes, I know, that `max_document_size` applies for all the POST/PUT requests, even for _config options. May be worth to fix that instead?

2. It seems like multipart requests are not handled. So the workaround of this limitation would be a multipart requests. Moreover, this will lead to quite strange cases. Assume we have ServerA and ServerB. ServerB limit documents by 2MiB size. ServerA has a document of 3MiB size and we want to replicate it. Because of the `single_max_doc_size` it won't, but we can add few attachments into it and it will: replicator uses multipart requests if document contains attachments for optimization reasons. So ServerA gets a document which it cannot update. If you uses deletion with preserve content, even admins cannot update such documents (but they should).

3. If server provides `_update` handlers, we still can create documents over the limit and, actually, update them without triggering 413 error.

4. Default may be quite too low. In 1.x we had 2GiB `max_document_size` which actually means "no limits". 1 MiB limit would be quite easily to hit and that would be a surprise error since even MongoDB allows docs to be up to 16MiB.

5. Sadly, we cannot configure it per database. Common case is when a single server is used for multiple applications. The goals of each are different as like as the data they operate. We would love to set limit as low as possible for public databases, but we have to respect needs of private ones. This makes this feature use quite complicated and actually, painful.

Resume: this feature is good indeed and helps to keep our CouchDB in good form, but the number of workaround and problems may not let users use it right. How can we deal with them?

src/chttpd_db.erl
```diff
@@ -1628,6 +1634,34 @@ bulk_get_open_doc_revs1(Db, Props, _, {DocId, Revs, Options}) ->
...
 
+remove_exceed_docs(DocArray0) ->
+    MaxSize = list_to_integer(config:get("couchdb",
+        "single_max_doc_size", "1048576")),
+    {ExceedDocs, DocsArray} = lists:splitwith(fun (Doc) ->
+        exceed_doc_size(Doc, MaxSize)
+    end, DocArray0),
+    ExceedErrs = lists:map (fun ({Doc}) ->
+        DocId = case couch_util:get_value(<<"_id">>, Doc) of
+            undefined -> couch_uuids:new();
```
kxepal: Why generate ID for new documents? That may confuse.

kxepal:
Some ideas:
1. Let `max_document_size` work only for document updates;
2. Add `require_content_length` option which enforces all the requests to specify Content-Length. This applies mostly for multipart body parts with document content. Or just always require that header without any additional options (generally, that's good idea);
3. Then, for `POST|PUT /db/doc` we can check `Content-Length` header and reject too large requests without parsing payload;
4. For multipart requests we can do the same (because p.2). Or just say, that document size is document size + sum(size of all the attachments), and then just again check `Content-Length` header, but without parsing payload. Transfer chunking request with multipart content doesn't works anyway (2295 issue iirc);
5. That was the lightweight logic. For `_bulk_docs` / `_update` handlers we'll have to use heavy logic based of determining document size from term format based on `exceed_doc_size/2`;

src/chttpd_db.erl
kxepal: I guess, without `?JSON_ENCODE` there is no since your size will be different for unicode data at least.

tonysun83:
@kxepal: Thanks for the quick reply and suggestions! I forgot about the multipart requests and realize it definitely complicates things. Let me re-think some things and come up with some new proposals/solutions.

src/chttpd_db.erl
tonysun83: Was on the fence of whether or not to generate an ID. I decided to leave it in for now. Will remove if I can't think of a valid reason later.

janl:
I like the idea here, but am also weary about adding a new config parameter that appears to be close to what we already have. Looking forward to what you come up with @tonysun83!

tonysun83:
@kxepal : The reason behind adding single_doc_max_size was because max_document_size was for all POST/PUT requests. For bulk_docs, it would look at the whole docs array and count that as a single document size rather than individual documents. If we only remove the limitation for all POST/PUT requests and move that logic over to the updates only, then having max_document_size should be sufficient.

Even though users SHOULD input content-length in their headers, I feel it might be a little restrictive in the general case. I think we should continue with parsing the payload as we normally do.

As for multi-parts, what do you think about using the same logic as update and not counting the attachments (normal and inline) at all? This still doesn't solve your replication example. I don't know enough about replication scenarios yet. ServerA is 2 MB max document size, ServerB contains a 3 MB document. We send it over and ServerA denies the update. Will replication still continue if we throw some error in this case? 

For now, if we're going to continue with the original max_document_size, I am okay not having it at DB  level. We can keep the default very high, and let the admin decide what he or she wants to do with the cluster.

kxepal:
@tonysun83 
I think it worth to make `max_document_size` to act as like as it name says or rename it to `max_http_payload_size` to avoid confusion.

> Even though users SHOULD input content-length in their headers, I feel it might be a little restrictive in the general case. I think we should continue with parsing the payload as we normally do.

Yes, agree, this even will break some stuff like browsers: I recall that Chrome may not send Content-Length for file upload or something. Though such restriction would simplify our file.

> As for multi-parts, what do you think about using the same logic as update and not counting the attachments (normal and inline) at all? 

Sounds as good idea and it will make a huge sense with external attachments storage feature. So to calculate document size we'll have to get it JSON form and strip `_attachments` field from it (we won't count stub info as well as it's something user cannot control, right?). 

> This still doesn't solve your replication example. I don't know enough about replication scenarios yet. ServerA is 2 MB max document size, ServerB contains a 3 MB document. We send it over and ServerA denies the update. Will replication still continue if we throw some error in this case? 

Yes, it will go on, just `doc_write_failures` counter will get incremented.

> For now, if we're going to continue with the original max_document_size, I am okay not having it at DB level. 

+1, since per DB configuration is another big feature. I just shared my observation about.

> We can keep the default very high, and let the admin decide what he or she wants to do with the cluster.

I'd thought about this over the day and I think we need a flag that turns on this document size restriction. Why? Because I don't think we would like to spend our resources to calculate document size all the time if our default limit is something like 2GiB - obliviously too high to hit it eventually. Let's say, our defaults are:
```
[couchdb]
max_document_size = 16777216  ; 16 MiB, what's fine for MongoDB, should be harmful for us.
use_max_document_size = false
```

The 16 MiB picked from a head, feel free to ignore. I think here is need to do some research about the most optimal value.

So, by default we: 
1) preserve semi-unlimited document size due to backward compatibility 
2) don't check document size at all and don't waste our cpu/memory for that

With release notes we strongly advice to set your limits up since we'll turn on them in future. The next step would be enable that limitation by default, and everyone are get aware about.

Sounds good for you, @tonysun83, @janl ?

tonysun83:
Sounds, good, I also like the use_max_document_size.

tonysun83:
@kxepal: I updated the code to reflect your suggestions above:

1) Keep max_document_size as actual limitiation
2) Add in use_max_document_size to see if we actually decide to verify the body
3) After parsing the doc, the body is stored in #doc.body. We check the size on this record field for comparison.
4) I moved the logic to verify the body size into updates only.
5) Multi_parts and _update also check #doc.body for size.
6) _bulk_doc extraction is only done when use_max_document_size is true.
7) Add testing for inline-attachments, multi-part/related, multi-part/form-data. (The multi-part tests were ridiculously painful because of all the boundary CRLF issues).

Let me know what you think.

src/chttpd.erl
```diff
@@ -858,6 +855,10 @@ error_info({error, {database_name_too_long, DbName}}) ->
         <<"At least one path segment of `", DbName/binary, "` is too long.">>};
 error_info({missing_stub, Reason}) ->
     {412, <<"missing_stub">>, Reason};
+error_info({request_entity_too_large, MaxSize}) ->
+    SizeBin = list_to_binary(integer_to_list(MaxSize)),
```
kxepal: Since we dropped <R16B03 support, it's safe to use integer_to_binary now.

src/chttpd_db.erl
```diff
@@ -1628,6 +1633,44 @@ bulk_get_open_doc_revs1(Db, Props, _, {DocId, Revs, Options}) ->
...
+            {[{id, DocId}, {error, <<"request_entity_too_large">>},
+                {reason, ?l2b(Reason)}]}
+            end, ExceedDocs),
+            {ExceedErrs, DocsArray};
+        false ->
+            {[], DocArray0}
+    end.
+
+exceed_doc_size(JsonBody, MaxSize) ->
+    size(term_to_binary(JsonBody)) > MaxSize.
```
kxepal: N.B.: need to test how `size(term_to_binary(JsonBody))` will be different from `size(?JSON_ENCODE(JsonBody))`. Quick checking shows, that it produces greater number, but I wonder how much deviation will be for something bigger.

kxepal:
That's awesome, @tonysun83! Thank you! Quickly looks good, I'll check it on the weekends.

P.S. Yes, multiparts are always painful /:

tonysun83:
Hmm, for a 1000 row document, there' seems to be significant difference between
```
size(term_to_binary(JsonBody)) 217688 
size(?JSON_ENCODE(JsonBody)) 164605
erlang:external_size(JsonBody) 282693
```

src/chttpd_db.erl
davisp: And we should be using erlang:external_size(Term) which doesn't waste resources trying to create the binary just to get a length.

davisp:
Couple things jump out at me here. First, this appears to be opening up a bit of a DOS vector in that the max body size is not hard coded to 4GiB without the ability for an admin to change that. I'd follow @kxepal's suggestion to add a max_http_body_size parameter that more specifically covers what max_document_size now covers.

Also, I would do away with trying to selectively enforce the document size based on a config parameter as well as just report errors for some documents in a _bulk_docs body. Just force the max_document_size check against everything and then throw an error if any doc in the array is too big (though include the id so the user knows which one it was).

I'd also added the check for _update and the like as well as mentioned.

For multipart/mime requests we should look at doing something similar but I'd implement it differently by crashing if we read more than max_document_size bytes from the socket (though we'll also want to come up with a plan on how that affects attachments). As mentioned we don't want to rely on the Content-Length header as that's not always set and not always knowable in advance (ie, think cat'ing a file to cURL upload or similar).

tonysun83:
Revisiting this.

@davisp : Any reason for multi-part/mime requests that we're going to the socket level? This commit https://github.com/apache/couchdb-chttpd/commit/5d7170c472d2ca72d4a0dd41eb7ae4434266be7d already performs a check based on max_document_size based and Content-Length. I'd go a step further (if the use_max_document_size option is set to true), and check the body size when there is no Content-Length available.

davisp:
I don't follow what you mean. All requests go to the socket if you go far enough.

tonysun83:
@davisp: I misunderstood you when you said "from the socket". For some reason I thought you meant that changes had to made at the mochiweb level. I think I get what you mean now...basically for multi-part requests, we're using ```receive_request_data``` to stream the request. As of right now, the current code waits for the entire stream to complete, and then checks the body to see if it surpasses max_document_size. What you're saying is that we shouldn't wait for the stream to complete, but rather fail mid stream if it exceeds max_document_size?

davisp:
Correct.

tonysun83:
@davisp: @kxepal: I've given this feature some more thought and I don't see how we can have two limits with only one config parameter. The current code hard codes a 4 MB limit for incoming body requests. In order to prevent DOS attacks, we should revert this back to a configurable limit (i.e ```max_document_size```). This would return us to the original behavior where ```max_document_size``` is really means max request size. 

But what we want is to have a scenario where a bulk_doc request can be say 16 MB, but each document individually does not exceed 2 MB. If we only have ```max_document_size```, that's not possible right?

mikerhodes:
@tonysun83 fwiw I think the suggestion is still to have the `max_http_body_size` new config setting in addition to making `max_document_size` work as it currently does, but sounds like we need to briefly sketch the config options and their interactions to get this straight :)

tonysun83:
Closing this as new PRs have been created

PR #115, opened 2016-04-20
Inject _design/_auth on auth_cache startup

iilyak:
We send start_listener event from init to make sure we call
ensure_auth_ddoc_exists as soon as we can. Rather then calling it on
listener restart.

COUCHDB-3000

### DIFF
--- a/src/chttpd_auth_cache.erl
+++ b/src/chttpd_auth_cache.erl
@@ -92,7 +92,8 @@ get_from_cache(UserName) ->
 %% gen_server callbacks
 
 init([]) ->
-    {ok, #state{changes_pid = spawn_changes(0)}}.
+    self() ! {start_listener, 0},
+    {ok, #state{}}.
 
 handle_call(_Call, _From, State) ->
     {noreply, State}.
@@ -116,8 +117,10 @@ handle_info({start_listener, Seq}, State) ->
 handle_info(_Msg, State) ->
     {noreply, State}.
 
-terminate(_Reason, #state{changes_pid = Pid}) ->
-    exit(Pid, kill).
+terminate(_Reason, #state{changes_pid = Pid}) when is_pid(Pid) ->
+    exit(Pid, kill);
+terminate(_Reason, _State) ->
+    ok.
 
 code_change(_OldVsn, #state{}=State, _Extra) ->
     {ok, State}.


src/chttpd_auth_cache.erl
```diff
@@ -95,7 +95,8 @@ get_from_cache(UserName) ->
 %% gen_server callbacks
 
 init([]) ->
-    {ok, #state{changes_pid = spawn_changes(0)}}.
+    self() ! {start_listener, 0},
```
kxepal: I don't have a lot experience, but I thought that sending a message to self on init is sort of a hack that may cause various issues like hanging and so. Is that true and safe to use here?

iilyak: There are five ways to deal with delayed initialization.

1. send a message to itself and handle it in handle_info. 
2. do gen_server:cast to itself. 
3. return `{ok, State, 0}` and implement `handle_info(timeout, ..`
4. use `proc_lib:start_link -> proc_lib:init_ack -> gen_server:enter_loop`

In this particular case we don't have any functions which use this gen_server (handle_call and handle_cast are empty and handle_info handles one specific case which wouldn't be a problem). So we don't care that someone would call us while we are not having changes_pid set. This means that `#4` is overkill for sure. Since we already handle `start_listener` event in handle_info it is easiest to use approach `#1`.
In cases when there are none empty handle_call or handle_cast and gen_server is anonymous I usually choose `#2` (i.e. `gen_server:cast(self(), init)`). Since nobody would be able to call us anyway because our gen_server doesn't have name and pid is not known to the potential caller. By the time we provide the pid to the potential caller init message would be in the process mailbox (because init is executed in the context of that process). Therefore the very first message gen_server will be processing would be `init` message. I would never use `#3` since it does rely on `timeout` atom which doesn't let you to clarify you intend.  `#4` is only useful for named processes. Because in this case you might get messages from potential callers before you place `init` message into process mailbox.

iilyak: I forgot to mention `#5` which I saw in some projects.

`erlang:send_after(10, self(), init).`

I wouldn't use it.

iilyak: Frankly I would choose `#2` in this case as well but I would have to refactor the code a little bit. This would slow down the review and would require writing a test suite.

kxepal: Thanks for detailed explanation! 
I'm agree with you that all cases that involves dependency on time shouldn't be used. The `#2` sounds good as it will require test suite (tests are good in any case), but I don't have any solid expertise here to judge.

davisp:
+1 I wouldn't bother with gen_server:cast/2 as its nothing more than self() ! {'$gen_cast', Msg} which is just adding unnecessary code.

iilyak:
`self() ! {'$gen_cast', Msg}` - is implementation detail of OTP which shouldn't be relied upon. The public interface of gen_server module is `gen_server:cast/2`.

davisp:
Right, I'm just saying there's no reason to prefer it over sending a message to self.

PR #116, opened 2016-04-23
restore 1.x behaviour: user docs in conflict cannot login

janl:


janl:
tests for this will be landing with https://github.com/apache/couchdb/pull/409

src/chttpd_auth_cache.erl
```diff
@@ -209,3 +206,18 @@ update_doc_ignoring_conflict(DbName, Doc, Options) ->
         throw:conflict ->
             ok
     end.
+
+validate_user_creds(nil) ->
+    nil;
+validate_user_creds(UserCreds) ->
+    couch_log:info("~n~n validate_user_creds: ~p~n~n", [UserCreds]),
```
KlausTrainer: No strong opinion, but I'd tend to rather use `debug` log level here.

KlausTrainer:
:+1:

src/chttpd_auth_cache.erl
kxepal: Agree. Logging on info will cause information leak though logs and systems they analyse in something more than developer environment.

kxepal: Also, why so much `~n`?

kxepal:
Except logging moment, +1. Nice catch!

src/chttpd_auth_cache.erl
janl: oh sorry, this is leftover debugging code, it’ll go away completely

rnewson:
might be good to have a config toggle on this, enabled by default.

kxepal:
@rnewson What's the case when we want to have such behaviour disabled? I don't think user will be happy if they would have specify one password or another depending on which conflict doc wins.

rnewson:
Well, I think we might turn it off for Cloudant so that customers aren't locked out of their accounts but that's a conversation we'll be having this week. Let's not pretend this is a desirable behaviour for users, just the best we can do easily.

janl:
New version without debugging statement and with config option

kxepal:
+1

src/chttpd_auth_cache.erl
```diff
@@ -209,3 +206,28 @@ update_doc_ignoring_conflict(DbName, Doc, Options) ->
         throw:conflict ->
             ok
     end.
+
+maybe_validate_user_creds(nil) ->
+    nil;
+maybe_validate_user_creds(UserCreds) ->
+    AllowConflictedUserDocs = config:get("chttpd_auth", "allow_conflicted_user_docs", "false"),
```
rnewson: use config:get_boolean

src/chttpd_auth_cache.erl
```diff
@@ -209,3 +206,28 @@ update_doc_ignoring_conflict(DbName, Doc, Options) ->
...
+    maybe_validate_user_creds(UserCreds, AllowConflictedUserDocs).
```
rnewson: then `AllowConfictedUserDocs andalso validate_user_creds(UserCreds)`

src/chttpd_auth_cache.erl
```diff
@@ -209,3 +206,28 @@ update_doc_ignoring_conflict(DbName, Doc, Options) ->
         throw:conflict ->
             ok
     end.
+
+maybe_validate_user_creds(nil) ->
```
rnewson: a `-spec` here would help readability, so we know what the possible results are supposed to be.

src/chttpd_auth_cache.erl
rnewson: enh, no, because validate_user_creds has an odd contract.

src/chttpd_auth_cache.erl
```diff
@@ -209,3 +206,28 @@ update_doc_ignoring_conflict(DbName, Doc, Options) ->
...
+maybe_validate_user_creds(UserCreds, "false") ->
+    UserCreds;
+maybe_validate_user_creds(UserCreds, "true") ->
+    validate_user_creds(UserCreds).
+
+validate_user_creds(nil) ->
+    nil;
+validate_user_creds(UserCreds) ->
+    case couch_util:get_value(<<"_conflicts">>, UserCreds) of
+    undefined ->
```
rnewson: indentation

rnewson:
I'd at least like to see the get_boolean comment addressed. The odd contract for the result might confound further simplification, though.

kxepal:
Oh, I should take a look on it closer. Thanks, @rnewson! Agree with your notes.

janl:
new version, uses `get_boolean`, added comment to clarify function behaviour

skipped spec because of odd function sig.

janl:
latest version should fix travis issues.

janl:
voilá

rnewson:
+1

janl:
thanks all!

PR #120, opened 2016-05-13
use shard suffix when generating _changes ETag

willholley:
In CouchDB 2.0, instance_start_time is always 0.
This means that when generating ETag values derived
from the database info object, the same ETags can be
incorrectly deemed valid between different database
instances with the metadata.

To avoid this we can incorporate the unique shard
suffix for the database instance when generating
the current ETag value.

COUCHDB-3017

### DIFF
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -90,7 +90,8 @@ handle_changes_req1(#httpd{}=Req, Db) ->
     "normal" ->
         T0 = os:timestamp(),
         {ok, Info} = fabric:get_db_info(Db),
-        Etag = chttpd:make_etag(Info),
+        Suffix = mem3:shard_suffix(Db),
+        Etag = chttpd:make_etag({Info, Suffix}),
         DeltaT = timer:now_diff(os:timestamp(), T0) / 1000,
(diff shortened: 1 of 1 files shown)


willholley:
I'd like to add tests for this but not quite sure where the appropriate place is. 

Also, this pattern is repeated  at https://github.com/apache/couchdb-global-changes/blob/203fb088ed81149108e64d18ffd9c3f5df154f97/src/global_changes_httpd.erl#L55 and https://github.com/apache/couchdb-couch/blob/b4295bfe58680c6a3d332a73ac6b061bd78b4db3/src/couch_httpd_db.erl#L123.

This makes me wonder whether changing the database info object to include the unique shard prefix would be a better solution.

wohali:
@davisp ping, we need an Erlang code review here, Jan and Bob are kinda busy and I'm out of my depth. Could you kindly help? Thanks :pig_nose: :cat: :cow2:

rnewson:
the #db record here is faked and not fully-populated (which is good, cause it would include disk_size and other things, which would make for a lousy etag).

I think the way forward is to fix this one and the global_changes one too. The instance where couch_db:get_db_info/1 is called is fine, since that's a real couchdb database (and thus instance_start_time does change for a delete/create)

PR #121, opened 2016-05-16
restore 1.x _users and _replicator behaviour

janl:


kxepal:
IIRC, views are also admin-only resources, right?

janl:
Since only admins can create views here, and we don’t install any views ourselves, we are good here, but good thinking that *if* we want to encourage people to use custom views, we should make those admin only as well.

janl:
@kxepal I’m currently looking at the JS tests that validate our auth behaviour, do you want to have a stab at making views admin-only here, too? Feel free to push on top of my branch

PR #123, opened 2016-05-27
Fix _node/$node/_system endpoint

eiri:
The endpoint is using the same handler as a backdoor interface, which only works for the requests to a local $node. For rpc calls to the remote nodes, as it's done on `/_node` endpoints, `handle_node_req` handler is getting `{badrpc,{'EXIT',normal}}` response and fails with `invalid_ejson` exception.

The change splits stats gathering from stats responding, allowing to use former on both `/_node/$node/_system` and backend's `/_system`

This closes COUCHDB-3027

### DIFF
--- a/src/chttpd_misc.erl
+++ b/src/chttpd_misc.erl
@@ -26,7 +26,8 @@
     handle_utils_dir_req/2,
     handle_uuids_req/1,
     handle_welcome_req/1,
-    handle_welcome_req/2
+    handle_welcome_req/2,
+    get_stats/0
 ]).
 
 -include_lib("couch/include/couch_db.hrl").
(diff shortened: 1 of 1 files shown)


src/chttpd_misc.erl
```diff
@@ -294,8 +295,9 @@ handle_node_req(#httpd{path_parts=[_, _Node, <<"_stats">>]}=Req) ->
     send_method_not_allowed(Req, "GET");
 % GET /_node/$node/_system
 handle_node_req(#httpd{method='GET', path_parts=[_, Node, <<"_system">>]}=Req) ->
-    Stats = call_node(Node, chttpd_misc, handle_system_req, [Req]),
-    chttpd:send_json(Req, Stats);
+    Stats = call_node(Node, chttpd_misc, get_stats, []),
+    EJSON = couch_stats_httpd:to_ejson(Stats),
+    chttpd:send_json(Req, EJSON);
```
iilyak: Nitpick. Here you specify module name while [on line 341](https://github.com/apache/couchdb-chttpd/pull/123/files#diff-938aabd9888322523915339d0185cf7aR341) you don't specify module name. Both forms would work, but it is better to not mix them without a reason.

iilyak:
Other than a small nitpick it looks good to me.

davisp:
+1

kxepal:
+1

PR #125, opened 2016-06-07
Don't expect #vacc as a response of end_delayed_json_response in multi_query_view

eiri:
When `multi_query_view` is calling `chttpd:end_delayed_json_response/1` it is passing it the response it received from query_view's `#vacc`, which means it'll get back a response record and not `#vacc`.

COUCHDB-3031

### DIFF
--- a/src/chttpd_view.erl
+++ b/src/chttpd_view.erl
@@ -33,8 +33,7 @@ multi_query_view(Req, Db, DDoc, ViewName, Queries) ->
         Acc1
     end, VAcc1, ArgQueries),
     {ok, Resp1} = chttpd:send_delayed_chunk(VAcc2#vacc.resp, "\r\n]}"),
-    {ok, Resp2} = chttpd:end_delayed_json_response(Resp1),
-    {ok, Resp2#vacc.resp}.
+    chttpd:end_delayed_json_response(Resp1).
 
 
 design_doc_view(Req, Db, DDoc, ViewName, Keys) ->


src/chttpd_view.erl
```diff
@@ -33,8 +33,7 @@ multi_query_view(Req, Db, DDoc, ViewName, Queries) ->
         Acc1
     end, VAcc1, ArgQueries),
     {ok, Resp1} = chttpd:send_delayed_chunk(VAcc2#vacc.resp, "\r\n]}"),
-    {ok, Resp2} = chttpd:end_delayed_json_response(Resp1),
-    {ok, Resp2#vacc.resp}.
+    chttpd:end_delayed_json_response(Resp1).
```
rnewson: returning Resp2#vacc.resp matches what design_doc_view returns and both are called only from handle_view_req.

I've reproduced this crash and confirmed this change fixes it, but it feels like something deeper is wrong and the true fix is elsewhere.

eiri: In `design_doc_view` the last thing we are calling `fabric:query_view/6` which _is_ returning vacc record. Here we are extracting response from vacc and then passing it through `chttpd:send_delayed_chunk/2` and  `chttpd:end_delayed_json_response/1`. i.e. we are passing around delayed_resp records, but expecting to get as a result vacc record as Resp2.

I'm basically reverting `multi_query_view` to how it is in dbcore and how it was in Couch prior [this commit](https://github.com/apache/couchdb-chttpd/pull/94). Why do you think this is incorrect fix?

davisp:
This looks like the right fix to me. +1

rnewson:
+1

iilyak:
@eiri: Did you check other functions altered in the commit you've mentioned?

- chttpd_db:all_docs_view
- design_doc_view:design_doc

eiri:
@iilyak `design_doc_view` used with POST and keys and that I did check, works fine. I haven't specially checked `all_docs_view`, but at first glance it looks like it should be good: vacc in -- vacc out. I think we'd notice if it'd be broken by now, it's not as esoteric as multi-query view

PR #126, opened 2016-06-07
Return error 410 on temporary view request

eiri:
The request for temporary view is currently returning "403 Forbidden" error, which is implying a permissions issue. Error "410 Gone" seems to fit better.

### DIFF
--- a/src/chttpd_view.erl
+++ b/src/chttpd_view.erl
@@ -77,4 +77,4 @@ handle_view_req(Req, _Db, _DDoc) ->
 
 handle_temp_view_req(Req, _Db) ->
     Msg = <<"Temporary views are not supported in CouchDB">>,
-    chttpd:send_error(Req, 403, forbidden, Msg).
+    chttpd:send_error(Req, 410, gone, Msg).


iilyak:
+1

kxepal:
Why bad? It's not bad since resource is actually not available anymore. `HTTP 410 Gone` would be more suitable.

kxepal:
Bad request actually means that user can fix it and get the response in the end. Here he cannot do that without patching CouchDB (:

eiri:
@kxepal user _can_ fix it by creating permanent view and then querying it ;) but I agree, 410 fits better, thought would require adding it to `chttpd:error_info/1`.

eiri:
Actually no, my bad, just calling `chttpd:send_error/4` should be fine.

davisp:
Actually a better solution to this would be to match the new temp view behavior in CouchDB single node behavior where we just create a design document and query that.

https://github.com/apache/couchdb-couch-mrview/blob/master/src/couch_mrview_http.erl#L143-L152

Theoretically that should be fairly straightforward to adapt to the cluster layer.

For background, the original CouchDB view engine used a temp file for building temp indexes that had no record in the actual database. This was difficult/ugly to support on a cluster so Cloudant just disabled them since they weren't a high demand feature. However when we rewrote the indexing system with couch_index/couch_mrview we changed the temp view behavior to just create a design document and query that which simplified the feature greatly. Now that that's been the behavior in CouchDB for so long on the cluster side we can do the same thing with the same behavior.

eiri:
Closing this, going to look into re-enabling  temp views instead.

PR #127, opened 2016-06-13
Don't try to decode params in multi_query_view

eiri:
Parameters in `multi_query_view` got decoded before parsing as part of reading POST's json body. This makes `parse_params` to skip double parsing and allows to use complex view keys in multi query payload.

COUCHDB-3031


eiri:
This depends on https://github.com/apache/couchdb-couch-mrview/pull/48

rnewson:
+1

davisp:
+1

PR #130, opened 2016-07-07
3049 fix sockname

jaydoane:
@rnewson, @iilyak

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -546,7 +546,7 @@ absolute_uri(#httpd{mochi_req=MochiReq, absolute_uri = undefined}, Path) ->
         undefined ->
             case MochiReq:get_header_value("Host") of
                 undefined ->
-                    {ok, {Address, Port}} = sockname(MochiReq:get(socket)),
+                    {ok, {Address, Port}} = inet:sockname(MochiReq:get(socket)),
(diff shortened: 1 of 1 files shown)


rnewson:
+1

kxepal:
So technically the fix of existed code should be:
```
sockname({ssl, Socket}) ->
    ssl:sockname(Socket).
```
?

kxepal:
/:

jaydoane:
Yes, I seem to have tested that code snippet on a non-mochiweb socket (without the tuple), which is utterly my bad :(

rnewson:
@kxepal The new code reflects exactly what couchdb has always done since 2012 (that code is in couch_httpd.erl).

kxepal:
@rnewson that doesn't means that it couldn't be improved (;

kxepal:
But it's too late anyway, so ok.

rnewson:
We'd want them to match, though. In my opinion the version in couch_httpd is quite clear.

rnewson:
@kxepal I hope it's also clear that the socket is _never_ a pid, that's the mistake, it's a port in the non-ssl case and a tuple, {ssl, blah, port, ...} if ssl.

kxepal:
@rnewson yes, you're right here.

PR #131, opened 2016-07-12
Don't end changes feed listener on heartbeat

rnewson:
There's no need to end and restart the changes feed listener on every
heartbeat, so don't do that.

COUCHDB-3054


kxepal:
+1

PR #133, opened 2016-07-21
In a view multi-query request, set view type for each query

nickva:
View type can be overridden if user explicitly sets reduce=false, for example.
This can happen for individual query.

Previously this didn't happen and view_type was set once at the top,  just
based on the view itself. For for example, a query like this:

```
{
  "queries": [{
    "include_docs": true,
    "reduce": false
  }]
}
```

would fail with:

```
{"error":"query_parse_error","reason":"`include_docs` is invalid for reduce"}
```

But it shouldn't fail because user explicitly disabled `reduce` for that one.

To fix, make sure to call `set_view_type` for each query's args.

Jira: COUCHDB-3070

### DIFF
--- a/src/chttpd_view.erl
+++ b/src/chttpd_view.erl
@@ -23,7 +23,8 @@ multi_query_view(Req, Db, DDoc, ViewName, Queries) ->
     ArgQueries = lists:map(fun({Query}) ->
         QueryArg = couch_mrview_http:parse_params(Query, undefined,
             Args1, [decoded]),
-        couch_mrview_util:validate_args(QueryArg)
+        QueryArg1 = couch_mrview_util:set_view_type(QueryArg, ViewName, Views),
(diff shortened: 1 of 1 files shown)


kxepal:
+1

nickva:
Updated pr with 2 tests. One checks when default query type is not overridden then `include_docs=true` should throw a validation exception. Other checks if user explicitly sets `reduce=false` and then query should succeed.

iilyak:
+1
```
    Running test function(s):
      chttpd_view:check_multi_query_reduce_view_overrides_test_/0
======================== EUnit ========================
chttpd_view:103: t_check_include_docs_throw_validation_error...[0.019 s] ok
chttpd_view:112: t_check_user_can_override_individual_query_type...[0.008 s] ok
=======================================================
  2 tests passed.
```

PR #137, opened 2016-08-19
Update handle_config_terminate API

iilyak:
Update handle_config_terminate API to match changes in apache/couchdb-config#11

COUCHDB-3102


davisp:
+1

PR #139, opened 2016-09-02
Drop logging level for "no record of user %s"

wohali:
This warning is tripped primarily when an admin is defined in local.ini
and has no associated _users document. This can happen fairly often in
development and testing setups, meaning that every other line in the
logfile is "no record of user admin".

The right thing to do here might be to check if the currently defined
user is a local admin and only complain in that situation, but presently
chttpd only directly depends on couch_auth_cache and to do so would 
require unencapsulating how couch_auth_cache handles ini file defined
admin users vs. authentication database users, which feels wrong.

Change suggested by @janl in #couchdb-dev IRC.


kxepal:
+1

PR #140, opened 2016-09-22
Handle disconnect when receiving body

iilyak:
When any error happen on an underlying socket mochiweb uses
exit(normal). Add catch for exit:normal and convert it to
exit({bad_request, <<"Incomplete">>}).

COUCHDB-3158


eiri:
Let's make the error "Invalid request body", to cover cases with `max_document_size` exceeded and because we seems to use more verbose forms on `bad_request` and +1

PR #141, opened 2016-09-27
Fix SSL defaults

rnewson:
Three properties had their defaults changed to a value that
couch_util:parse_terms throws up on.

`"nil"` was changed to `undefined` in COUCHDB-2561 as part of normalising
the config API. The change from list to atom was a mistake.

COUCHDB-3162

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -61,9 +61,9 @@ start_link(http) ->
 
 start_link(https) ->
     Port = config:get("ssl", "port", "6984"),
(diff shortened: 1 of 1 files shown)


nickva:
+1

couch_util:parse_term has a clause for binaries, should be change it to handle atoms as well?

rnewson:
no, I don't think so.

nickva:
Ah makes sense, it would be inconsistent.  It takes external-like things (string and binaries), atoms don't fit there.

PR #142, opened 2016-09-27
Fix eventsource timeout handling

rnewson:
COUCHDB-3164


abhishiv:
Just hit this bug. 

On jira i see this path is scheduled for 2.1.0. Any chances to slip this in with the next 2.0.0-rc?

PR #143, opened 2016-10-04
Add max_document_size checking for multipart PUT requests

nickva:
Previously multipart/related PUT requests didn't check maximum request sizes.

This commit checks content-length and compares that with the maximum.

This means keeping the current "semantics" of max_document_size which actually
means "max request size". But this makes the check more efficient and can
be done earlier in request processing time.

PR depends on https://github.com/apache/couchdb-couch/pull/201 make sure to
merge that one first if accepted.

Jira: COUCHDB-3174


davisp:
+1

PR #147, opened 2016-10-17
Make auth handlers configurable.

dmunch:
Unit tests pass, dev cluster runs. Plus I'm able to use a modified version of couch_jwt_auth.

### DIFF
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -490,10 +490,8 @@ set_auth_handlers() ->
 % - this only works for modules with exactly two underscores in their name
 % - is not very robust (a space after the ',' is assumed) 
 auth_handler_name(SpecStr) ->
-    case couch_util:parse_term(SpecStr) of
-    {ok, {_, Fun}} ->
-        hd(string:tokens(atom_to_list(Fun), "_"))
-    end.
+    {ok, {_, Fun}} = couch_util:parse_term(SpecStr),
+    hd(string:tokens(atom_to_list(Fun), "_")).
 
 authenticate_request(Req) ->
     {ok, AuthenticationFuns} = application:get_env(chttpd, auth_handlers),
--- a/src/chttpd.erl
+++ b/src/chttpd.erl
@@ -117,6 +117,8 @@ start_link(Name, Options) ->
          end,
     ok = couch_httpd:validate_bind_address(IP),
 
+    set_auth_handlers(),
+
     Options1 = Options ++ [
         {loop, fun ?MODULE:handle_request/1},
         {name, Name},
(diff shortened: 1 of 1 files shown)


src/chttpd.erl
```diff
@@ -464,12 +466,37 @@ extract_cookie(#httpd{mochi_req = MochiReq}) ->
...
+% chttpd_auth:default_authentication_handler: default
+% chttpd_auth_cookie_authentication_handler: cookie
+% couch_http_auth:proxy_authentication_handler: proxy
+%
+% couch_http:auth_handler_name can't be used here, since it assumes the name 
+% of the auth handler to be the 6th token split by [\\W_]
+% - this only works for modules with exactly two underscores in their name
+% - is not very robust (a space after the ',' is assumed) 
+auth_handler_name(SpecStr) ->
+    case couch_util:parse_term(SpecStr) of
```
rnewson: a case statement with only one clause seems odd.

rnewson:
just one comment, otherwise looks good to me. will merge after that case clause is looked at. I suspect we just want an assertive `{ok... = parse` there, since there's no recovery clause.

dmunch:
Agreed. Resulted from a stripped-down copy-and-pasted method. Replaced it with the assertive like you suggested.

kocolosk:
Nice @dmunch. This remains a somewhat janky part of our codebase and I think you were smart to do this in a minimally-invasive way. 👍

rnewson:
@dmunch thank you for your contribution, I've merged this work so please close the PR. This didn't happen automatically as I squashed your two commits together and removed some trailing whitespace introduced in your first commit.

dmunch:
Great! Thanks for merging!

ChiragMoradiya:
May I know, when this will be released?

dholth:
Thank you @dmunch ! It works for me in CouchDB 2.1.1. I added the following lines to `local.ini` to enable proxy authentication, and to check proxy before cookie. The default `{couch_httpd_auth}` configuration is in the wrong section.

```
[chttpd]
authentication_handlers = {couch_httpd_auth, proxy_authentication_handler}, {couch_httpd_auth, cookie_authentication_handler}, {couch_httpd_auth, default_authentication_handler}

[couch_httpd_auth]
secret = (32 random hex digits)
proxy_use_secret = true
```

wohali:
I believe we recently fixed this in our documentation (post-2.1.1):

https://github.com/apache/couchdb-documentation/pull/220

dholth:
The "latest" documentation is okay, but the correct location of proxy_use_secret is in a different part of the documentation. Also, proxy_use_secret is not present in the web configuration interface by default, and I don't see documentation of its default value.

PR #148, opened 2016-11-02
handle timeout in _revs_diff

mayya-sharipova:
when fabric workers timeout,
fabric:get_missing_revs returns {error,timeout}
this handles timeout errors

COUCHDB-3220

### DIFF
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -532,11 +532,16 @@ db_req(#httpd{path_parts=[_,OP]}=Req, _Db) when ?IS_ALL_DOCS(OP) ->
 db_req(#httpd{method='POST',path_parts=[_,<<"_missing_revs">>]}=Req, Db) ->
     chttpd:validate_ctype(Req, "application/json"),
     {JsonDocIdRevs} = chttpd:json_body_obj(Req),
-    {ok, Results} = fabric:get_missing_revs(Db, JsonDocIdRevs),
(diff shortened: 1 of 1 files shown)


src/chttpd_db.erl
```diff
@@ -532,31 +532,40 @@ db_req(#httpd{path_parts=[_,OP]}=Req, _Db) when ?IS_ALL_DOCS(OP) ->
 db_req(#httpd{method='POST',path_parts=[_,<<"_missing_revs">>]}=Req, Db) ->
     chttpd:validate_ctype(Req, "application/json"),
     {JsonDocIdRevs} = chttpd:json_body_obj(Req),
-    {ok, Results} = fabric:get_missing_revs(Db, JsonDocIdRevs),
-    Results2 = [{Id, couch_doc:revs_to_strs(Revs)} || {Id, Revs, _} <- Results],
-    send_json(Req, {[
-        {missing_revs, {Results2}}
-    ]});
+    case fabric:get_missing_revs(Db, JsonDocIdRevs) of
+        {error, timeout} ->
```
rnewson: could we not do `{error, Reason} -> chttpd:send_error(Req, Reason)` here and below?

rnewson: I mean, why only handle the specific timeout error? we only expect an {ok, Results} or an {error, Reason} result from fabric:get_missing_revs, anything other than those two would be unexpected (and, in the OTP spirit, we would write error handling code for them), but it seems reasonable to handle all {error, Reason} results.

rnewson:
+1

PR #149, opened 2016-11-04
HTTP code 500 instead of 400 for invalid key during document creation

mayya-sharipova:
The wrong status code is returned when creating or updating a document
with an invalid key.
Returns:
{u'reason': u'Bad special document member: _invalid_key', u'error':
u'doc_validation'}
HTTPError: 500 Server Error: Internal Server Error
Expected: 400 Bad Request error

COUCHDB-3222


rnewson:
+1

PR #150, opened 2016-11-11
Include user_ctx in db open options

eiri:
Previously original `Db` record was kept on `filter_fun` tuple in `#changes_args` and passed to the filters. Now we are throwing it away and including `Db` record build in fabric, which has `user_ctx` stripped from it. As a result `UserCtx` object passed to the filter functions in ddoc is missing user's name and role.

The fix includes original `user_ctx` into db open parameters, so `Db` record build in fabric will have it properly set.

COUCHDB-3233

### DIFF
--- a/src/chttpd_db.erl
+++ b/src/chttpd_db.erl
@@ -83,7 +83,8 @@ handle_changes_req(#httpd{path_parts=[_,<<"_changes">>]}=Req, _Db) ->
 handle_changes_req1(#httpd{}=Req, Db) ->
     #changes_args{filter=Raw, style=Style} = Args0 = parse_changes_query(Req),
     ChangesArgs = Args0#changes_args{
-        filter_fun = couch_changes:configure_filter(Raw, Style, Req, Db)
+        filter_fun = couch_changes:configure_filter(Raw, Style, Req, Db),
+        db_open_options = [{user_ctx, Db#db.user_ctx}]
     },
     Max = chttpd:chunked_response_buffer_size(),
     case ChangesArgs#changes_args.feed of


src/chttpd_db.erl
```diff
@@ -83,7 +83,8 @@ handle_changes_req(#httpd{path_parts=[_,<<"_changes">>]}=Req, _Db) ->
 handle_changes_req1(#httpd{}=Req, Db) ->
     #changes_args{filter=Raw, style=Style} = Args0 = parse_changes_query(Req),
     ChangesArgs = Args0#changes_args{
-        filter_fun = couch_changes:configure_filter(Raw, Style, Req, Db)
+        filter_fun = couch_changes:configure_filter(Raw, Style, Req, Db),
+        db_open_options = [{user_ctx, Db#db.user_ctx}]
```
iilyak: For consistency reasons I propose to extract user_ctx from Req rather than from Db. Even though [they are the same](https://github.com/apache/couchdb-chttpd/blob/master/src/chttpd_db.erl#L311)

eiri: To be honest I pulled it from Db exactly for consistency reason. First, this is what we've been doing previously, i.e. passing Db and _its_ user context and propagating UserCtx. Second, we are opening db in fabric with db open options, so for me it seems more natural to do that with the same options that was used to open db here, rather than pull context from request record. Plus if (when) we move to Cowboy it'll be one less place to chase and change.

iilyak: +1

iilyak:
+1

PR #152, opened 2016-12-12
Fix missing ampersand in rewritten queries

Murgeye:
Fix related to issue https://issues.apache.org/jira/browse/COUCHDB-3254


rnewson:
nice find, +1.

PR #155, opened 2017-02-09
Mock config module in tests

iilyak:
The tests in chttpd_db_bulk_get_test do not start config app.
The change in https://github.com/apache/couchdb-couch/pull/226
introduces call to config app.
Make sure we mock config so the tests pass.

COUCHDB-3293

### DIFF
--- a/test/chttpd_db_bulk_get_test.erl
+++ b/test/chttpd_db_bulk_get_test.erl
@@ -19,6 +19,7 @@
 
 
 setup() ->
+    mock(config),
     mock(chttpd),
     mock(couch_epi),
     mock(couch_httpd),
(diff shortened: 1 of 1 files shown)


nickva:
+1

```
chttpd_error_info_tests: error_info_test (module 'chttpd_error_info_tests')...[0.010 s] ok
=======================================================
  All 136 tests passed.
==> rel (eunit)
==> db-chttpd (eunit)
```

nickva:
Note: I couldn't reproduce this error locally. However I have seen similar failure when working on couch pr resulting from config being in the path of basic doc operations. The fix was similar to here so and tests still pass with the pr.

PR #157, opened 2017-03-13
Allow limiting maximum document body size

nickva:
This is the HTTP layer and some tests. The actual checking is done in couch
application's from_json_obj/1 function.

If a document is too large it will return a 413 response code. The error reason
will be the document ID. The intent is to help users identify the document if
they used _bulk_docs endpoint. It will also help replicator skip over documents
which are too large.

COUCHDB-2992


nickva:
Tests are failing because it needs changes from https://github.com/apache/couchdb-couch/pull/235

sagelywizard:
Doesn't look like this validates unclustered _update handler docs in `couch_mrview_show:handle_doc_update_req/3`

sagelywizard (approved):
But that doesn't require any changes for this repo. 🦄

